PR #136, opened 2022-02-16
A batch of bug fixes in the install command.

3asm:


src/ostorlab/cli/agent/delete/delete.py
```diff
@@ -5,25 +5,25 @@
...
     """CLI command to list installed agents."""
     docker_client = docker.from_env()
     images = docker_client.images.list()
-    agent_container_name = agent_key.replace('/', '_')
+    agent_container_name = agent.replace('/', '_')
     deleted = False
     for im in images:
         for t in im.tags:
-            if t.split(':')[0] == agent_container_name:
+            if t.split(':')[0] in (agent_container_name, f'ostorlab.store/agents/{agent_container_name}'):
```
deadly-panda, line 26: I don't understand the purpose of : (agent_container_name, f'ostorlab.store/agents/{agent_container_name}')

src/ostorlab/cli/agent/delete/delete.py
```diff
@@ -32,4 +32,4 @@ def delete(agent_key: str, agent_version_regex: Optional[str] = None) -> None:
                     deleted = True
 
     if deleted is False:
-        console.error(f'No agent matching [bold white]{agent_key}[/] were found')
+        console.error(f'No agent matching [bold white]{agent}[/] were found')
```
deadly-panda, line 35: was found*

src/ostorlab/cli/agent/delete/delete.py
3asm: The agents when they are installed, produce 2 images, one tagged from the agent name and one with the remote name. Hence we need to delete both.

PR #137, opened 2022-02-16
Multiple fixes

amine3:


src/ostorlab/cli/scan/list/list.py
```diff
@@ -28,8 +28,6 @@ def list_scans(ctx: click.core.Context, page: int, elements: int) -> None:
             console.success('Scans listed successfully.')
             columns = {'Id': 'id',
                        'Application': 'application',
```
3asm, line 30: Application needs to be replaced by asset. It also still missing.

3asm (changes requested):

3asm (approved):

PR #168, opened 2022-02-24
Reportvuln log

deadly-panda:


3asm:
Should not force logging as part of the SDK.

PR #176, opened 2022-02-28
Describe all vulnz of a single scan.

3asm:


amine3 (changes requested):
Missing unit test, Could you please add one

amine3 (approved):
LGTM

PR #180, opened 2022-03-02
Domaine name proto message.

deadly-panda:


src/ostorlab/agent/message/proto/v3/asset/domain_name/domain_name.proto
```diff
@@ -0,0 +1,7 @@
+syntax = "proto2";
```
deadly-panda, line 1: domaine_name.proto in a dir called domaine_name, not directly in asset/ correct ?

3asm (approved):

PR #196, opened 2022-03-08
Api requests for cloud scan run

deadly-panda:


src/ostorlab/runtimes/definitions.py
```diff
@@ -137,6 +137,8 @@ def to_raw_proto(self) -> bytes:
 class AgentGroupDefinition:
     """Data class holding the attributes of an agent."""
     agents: List[AgentSettings]
+    name: Optional[str] = ''
```
3asm, line 140: None.

src/ostorlab/runtimes/definitions.py
```diff
@@ -162,7 +166,17 @@ def from_yaml(cls, group: io.FileIO):
                             for p
                             in agent.get('open_ports', [])]
             )
-
             agent_settings.append(agent_def)
 
-        return cls(agent_settings)
+        name = agent_group_def.get('name', '')
```
3asm, line 171: This is not needed, can / should be removed.

3asm (changes requested):

PR #198, opened 2022-03-09
Cli cloud list vulnz

mohsinenar:


src/ostorlab/apis/runners/authenticated_runner.py
```diff
@@ -133,11 +132,12 @@ def execute(self, request: api_request.APIRequest) -> Dict:
             console.warning('No authentication credentials were provided.')
 
         response = self._sent_request(request, headers)
+        data = response.json()
```
3asm, line 135: I don't see the point of this change?

src/ostorlab/cli/auth/revoke/revoke.py
```diff
@@ -2,7 +2,7 @@
 
 import logging
 
-from ostorlab.apis.runners import authenticated_runner
+from ostorlab.apis.runners import authenticated_runner , runner
```
3asm, line 5: We prefer having import over separate lines.

3asm (approved):

PR #213, opened 2022-03-14
Cloud runtime vulnz dump

mohsinenar:


setup.cfg
```diff
@@ -61,6 +61,7 @@ install_requires =
     sqlalchemy
     semver
     markdownify
+    jsonlines
```
3asm, line 64: Havine a dependency for jsonlines is an overkill. We need to keep deps to the minimum.

3asm (changes requested):

src/ostorlab/cli/dumpers.py
```diff
@@ -1,18 +1,17 @@
 """Module responsible for dumping data to different formats."""
 
 import abc
-import json
 import csv
-from typing import Dict
-
+import json
+import typing
```
3asm, line 6: typing is an exception to class import per style guide. You can use from typing import List

src/ostorlab/runtimes/local/runtime.py
```diff
@@ -550,3 +553,29 @@ def describe_vuln(self, scan_id: int, vuln_id: int):
             console.success('Vulnerabilities listed successfully.')
         except sqlalchemy.exc.OperationalError:
             console.error('Vulnerability / scan not Found.')
+
+    def dump_vulnz(self, scan_id: int, output: str, output_format: str):
```
3asm, line 557: I don't see this added to the runtime base class.

3asm, line 557: I don't see this added to the runtime base class.

src/ostorlab/runtimes/local/runtime.py
```diff
@@ -550,3 +553,29 @@ def describe_vuln(self, scan_id: int, vuln_id: int):
...
+        for vulnerability in vulnerabilities:
+            vuln = {
+                'id':vulnerability.id,
+                'risk_rating': vulnerability.risk_rating.value,
+                'cvss_v3_vector': vulnerability.cvss_v3_vector,
+                'title': vulnerability.title,
+                'short_description': vulnerability.short_description
+            }
+            vulnz_list.append(vuln)
+        if output_format == 'csv':
```
3asm, line 576: This is not a great api design since every time we add format, we would need to change the API to add it.

A better impelement is that all API return a list in a specific format and that list get passed to the selected dumper class, this way it is agnostic to the format.

3asm, line 576: This is not a great api design since every time we add format, we would need to change the API to add it.

A better impelement is that all API return a list in a specific format and that list get passed to the selected dumper class, this way it is agnostic to the format.

3asm (changes requested):

src/ostorlab/cli/dumpers.py
```diff
@@ -57,10 +57,9 @@ def dump(self) -> None:
             FileNotFoundError: in case the path or file name are invalid.
         """
         if not self.output_path.endswith('.csv'):
-            self.output_path+= '.csv'
-        with open(self.output_path , 'w', encoding='utf-8') as outfile:
-            csv_writer = csv.DictWriter(outfile, fieldnames = FIELDNAMES)
+            self.output_path += '.csv'
+        with open(self.output_path, 'a', encoding='utf-8') as outfile:
```
3asm, line 61: Why change the mode to append?

tests/cli/vulnz/describe/test_describe_vulnz.py
```diff
@@ -112,4 +112,4 @@ def testOstorlabCloudRuntimeScanVulnzDescribeCLI_whenScanNotFound_showNotFoundEr
 
     result = runner.invoke(rootcli.rootcli, ['vulnz', '--runtime', 'cloud', 'describe', '-scan-id=502152'])
 
-    assert 'Scan can-id=502152 not found' in result.output
+    assert 'Vulnerability / scan not Foun' in result.output
```
3asm, line 115: Typo in Found. Also all should be lowercase.

3asm (changes requested):

tests/cli/vulnz/describe/test_describe_vulnz.py
mohsinenar, line 115:
```suggestion
    assert 'Vulnerability/scan not found.' in result.output
```

3asm (approved):

PR #228, opened 2022-03-26
Add markdown rendering to search description.

3asm:


src/ostorlab/cli/agent/search/search.py
```diff
@@ -34,7 +36,7 @@ def search_cli(keyword: str) -> None:
                 version = versions[0].get('version')
                 description = versions[0].get('description')
                 in_selectors = ', '.join(versions[0].get('inSelectors'))
```
amine3, line 38: why do have a space here?

amine3 (changes requested):

PR #236, opened 2022-03-30
Add network to litelocal runtime and add it to ostorlab CLI.

3asm:
Make the debug parameters hidden as only used for testing.


amine3 (approved):
LGTM

PR #247, opened 2022-04-05
Agent persist mock to help with testing persist mixin.

3asm:


src/ostorlab/testing/agent.py
```diff
@@ -35,5 +35,32 @@ def mq_send_message(key, message):
     emitted_messages = []
 
 
+@pytest.fixture(scope='function')
```
deadly-panda, line 38: i think u can remove the function scope, it s the default one.

deadly-panda (approved):

PR #252, opened 2022-04-08
Fix geolocation proto types.

deadly-panda:


3asm (changes requested):
This is only fixing ipv4.

deadly-panda:
> This is only fixing ipv4.

The ipv6 is already correct.

3asm (approved):

PR #255, opened 2022-04-15
Add job to deploy ostorlab docker image.

mohsinenar:


.github/workflows/python-publish.yml
```diff
@@ -43,3 +43,29 @@ jobs:
...
+          username: ${{ secrets.store_user }}
+          password: ${{ secrets.store_password }}
+
+      - name: Build image and push to ostorlab Container Registry
+        uses: docker/build-push-action@v2
+        with:
+          # relative path to the place where source code with Dockerfile is located
+          context: .
+          tags: |
+            ostorlab.store/test_project/test_agent:${{github.ref_name}}
```
3asm, line 69: These links are not correct.

3asm: Does refname have the v or is is without it?

Dockerfile
```diff
@@ -2,7 +2,7 @@ FROM python:3.9-bullseye as base
 FROM base as builder
 RUN mkdir /install
 WORKDIR /install
-RUN pip install --prefix=/install ostorlab
+RUN pip install --prefix=/install ostorlab -U
```
3asm, line 5: This should be built from the code, otherwise it will pick an older version from pypi.

3asm (changes requested):

3asm (approved):

3asm (approved):

PR #277, opened 2022-05-08
Add detail to protos

amine3:


src/ostorlab/agent/message/proto/v3/fingerprint/domain_name/library/library.proto
```diff
@@ -7,4 +7,5 @@ message Message {
   required string library_name = 2;
   optional string library_version = 3;
   required string library_type = 4;
+  required string detail = 5;
```
deadly-panda, line 10: I think agent what web uses this proto as an out selector, remember to add this required field to the agent logic & unit tests!

deadly-panda (approved):

3asm (approved):

PR #299, opened 2022-07-04
add domain_name service protobuf

mohsinenar:


src/ostorlab/agent/message/proto/v3/fingerprint/domain_name/service/library/library.proto
```diff
@@ -1,6 +1,6 @@
 syntax = "proto2";
 
-package v3.fingerprint.domain_name.library;
+package v3.fingerprint.domain_name.service.library;
 
 message Message {
   required string domain_name = 1;
```
3asm, line 6: Missing port and schema.

3asm (changes requested):

src/ostorlab/agent/message/proto/v3/fingerprint/domain_name/service/library/library.proto
```diff
@@ -1,11 +1,13 @@
...
   required string library_name = 2;
   optional string library_version = 3;
   required string library_type = 4;
   required string detail = 5;
+  required string port = 6;
```
3asm, line 11: 1, 2, 3 should be the same as the asset.
So they should b.

3asm (changes requested):

3asm (approved):

PR #306, opened 2022-07-14
Fix long image name breaking docker config name

amine3:


src/ostorlab/runtimes/lite_local/agent_runtime.py
```diff
@@ -170,6 +171,7 @@ def create_settings_config(self) -> docker.types.ConfigReference:
         """
         agent_instance_settings_proto = self.agent.to_raw_proto()
         config_name = f'config_settings_{self.image_name}_{self.runtime_name}'
+        config_name = hashlib.md5(config_name.encode()).hexdigest()
```
3asm, line 174: This needs to be done for all runtimes.

3asm (changes requested):

src/ostorlab/runtimes/lite_local/agent_runtime.py
```diff
@@ -193,6 +195,7 @@ def create_definition_config(self) -> docker.types.ConfigReference:
         """
         agent_definition = self._docker_client.images.get(self.agent.container_image).labels.get('agent_definition')
         config_name = f'config_definition_{self.image_name}__{self.runtime_name}'
+        config_name = hashlib.md5(config_name.encode()).hexdigest()
```
3asm, line 198: Please inline the string in here, no need to put in a variable that gets reassigned, it looks incorrect.

3asm (changes requested):

3asm (approved):

PR #308, opened 2022-07-19
Add email field to IP whois protobuff message.

deadly-panda:


src/ostorlab/agent/message/proto/v3/asset/ip/v4/whois/whois.proto
```diff
@@ -13,6 +13,7 @@ message Contact {
   optional string name = 1;
   optional string kind = 2;
   optional string address = 3;
+  optional string email = 4;
```
3asm, line 16: Do we have email in the contact ?

3asm (changes requested):

3asm (approved):

PR #314, opened 2022-07-27
Implement arch check

jamu85:
Raises an error if the runtime or agents installs are started on unsupported machine architecture.
As the issue is related to the docker requirements, I implemented the check in docker_requirements_checker.py
New supported architectures need to be added to the list of supported architectures.

The architecture is checked with a return value that's based on the kernel implementation of the uname(2)
 system call. So it might be necesarry to handle the same arch with various strings e.g. linux returns x86_64
 or AMD64 on windows.

On mac os, uname returns x86 even on arm64 if the process calling it is running via rosetta. Better parse for ARM
in platform.version() to determine the arch on mac os

### DIFF
--- a/src/ostorlab/cli/agent/install/install.py
+++ b/src/ostorlab/cli/agent/install/install.py
@@ -21,7 +21,6 @@ def install(agent: str, version: str = '') -> None:
         console.error('Docker is not installed.')
         raise click.exceptions.Exit(2)
     elif not docker_requirements_checker.is_sys_arch_supported():
-        console.info("install call")
         console.error('System architecture is not supported.')
         raise click.exceptions.Exit(2)
     elif not docker_requirements_checker.is_user_permitted():
--- a/src/ostorlab/cli/agent/install/install.py
+++ b/src/ostorlab/cli/agent/install/install.py
@@ -20,6 +20,10 @@ def install(agent: str, version: str = '') -> None:
     if not docker_requirements_checker.is_docker_installed():
         console.error('Docker is not installed.')
         raise click.exceptions.Exit(2)
+    elif not docker_requirements_checker.is_sys_arch_supported():
+        console.info("install call")
+        console.error('System architecture is not supported.')
+        raise click.exceptions.Exit(2)
     elif not docker_requirements_checker.is_user_permitted():
         console.error('User does not have permissions to run docker.')
         raise click.exceptions.Exit(2)
--- a/src/ostorlab/cli/docker_requirements_checker.py
+++ b/src/ostorlab/cli/docker_requirements_checker.py
@@ -1,6 +1,7 @@
 """Check if requirements for running docker are satisfied."""
 
 import docker
+import platform
 import sys
 from docker import errors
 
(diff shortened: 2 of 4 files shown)


src/ostorlab/cli/agent/install/install.py
```diff
@@ -20,6 +20,9 @@ def install(agent: str, version: str = '') -> None:
     if not docker_requirements_checker.is_docker_installed():
         console.error('Docker is not installed.')
         raise click.exceptions.Exit(2)
+    elif not docker_requirements_checker.is_sys_arch_supported():
```
3asm, line 23: Nice job figuring out where to add the checks.

src/ostorlab/cli/docker_requirements_checker.py
```diff
@@ -18,6 +19,24 @@ def is_docker_installed() -> bool:
...
+    """Checks if the systems cpu architecture is supported
+       The architecture is checked with a return value that's based on the kernel implementation of the uname(2)
+       system call. So it might be necesarry to handle the same arch with various strings e.g. linux returns x86_64
+       or AMD64 on windows.
+    Returns:
+        True if the architecture is supported, else False
+    """
+    supported_arch_types = ["x86_64", "AMD64"]
+    if sys.platform == 'darwin' and 'ARM' in platform.version():
+        """On mac os, uname returns x86 even on arm64 if the process calling it is running via rosetta. We parse for ARM
```
3asm, line 32: This should be a comment style, I think the linter will not pass this. As FYI, we apply Google style guide in the code base for consistency : https://google.github.io/styleguide/pyguide.html

src/ostorlab/cli/docker_requirements_checker.py
```diff
@@ -18,6 +19,24 @@ def is_docker_installed() -> bool:
...
 
+def is_sys_arch_supported() -> bool:
+    """Checks if the systems cpu architecture is supported
+       The architecture is checked with a return value that's based on the kernel implementation of the uname(2)
+       system call. So it might be necesarry to handle the same arch with various strings e.g. linux returns x86_64
+       or AMD64 on windows.
+    Returns:
+        True if the architecture is supported, else False
+    """
+    supported_arch_types = ["x86_64", "AMD64"]
```
3asm, line 30: This is better put as a const at the top of the file for readability's sake.

src/ostorlab/cli/docker_requirements_checker.py
```diff
@@ -18,6 +19,24 @@ def is_docker_installed() -> bool:
...
+    if sys.platform == 'darwin' and 'ARM' in platform.version():
```
3asm, line 31: Great catch!

3asm (changes requested):

jamu85 (commented):
Should be according to your request now

3asm (approved):

jamu85:
I could fix the linter issues, however, I have no idea how to solve the ssh key task for testing. Can this pr be closed?

3asm:
I don't understand either why the SSH is failing. I am suspecting some permission bit. I believe we no longer need the SSH in the flow, but I need to double-check.

PR #324, opened 2022-08-05
Fix all protos build from src.

3asm:


src/ostorlab/agent/message/proto/README.md
```diff
@@ -1,5 +1,5 @@
-It important to build the file from the proto folder to maintain the package structure. A onliner to build all protos
-in the v3 filder:
+It is important to build the file from the src folder to maintain the package structure. A onliner to build all protos
+in the `src` folder:
 ```shell
 find v3 -name '*.proto' | xargs -I {} protoc --python_out=. {}
```
deadly-panda, line 4: The 'find v3' wont compile common protos.
I think u need to specify the -path option because v3 dir is not directly under src dir.

deadly-panda (approved):

PR #327, opened 2022-08-08
Parallelize agent start with a thread pool executor.

3asm:


src/ostorlab/runtimes/local/runtime.py
```diff
@@ -330,8 +331,11 @@ def _check_agents_healthy(self):
 
     def _start_agents(self, agent_group_definition: definitions.AgentGroupDefinition):
         """Starts all the agents as list in the agent run definition."""
-        for agent in agent_group_definition.agents:
-            self._start_agent(agent, extra_configs=[])
+        with futures.ThreadPoolExecutor() as executor:
```
amine3, line 334: Do we need to set a max_workers?

amine3 (approved):

src/ostorlab/runtimes/local/runtime.py
3asm: I prefer keeping the default to scale with the number of cores.

PR #332, opened 2022-08-15
Opentelemetry mixin impelmentation adding function tracing

deadly-panda:
This PR is a follow up to #329. It presents the OpenTelemetry mixin, that will be used by the Agents
to add traces of all their methods, processing emiting and even cleanup. The implementation is basic for now, adding only the name of the agent, the operation, and the selector in case of emit and process operations.

### DIFF
--- a/src/ostorlab/runtimes/lite_local/agent_runtime.py
+++ b/src/ostorlab/runtimes/lite_local/agent_runtime.py
@@ -240,7 +240,7 @@ class AgentRuntime:
         self.agent.healthcheck_host = HEALTHCHECK_HOST
         self.agent.healthcheck_port = HEALTHCHECK_PORT
         self.agent.redis_url = self.redis_url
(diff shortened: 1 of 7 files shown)


src/ostorlab/runtimes/lite_local/agent_runtime.py
```diff
@@ -240,7 +240,7 @@ def update_agent_settings(self) -> None:
         self.agent.healthcheck_host = HEALTHCHECK_HOST
         self.agent.healthcheck_port = HEALTHCHECK_PORT
         self.agent.redis_url = self.redis_url
-        self.agent.tracing_collector_url = self.tracing_collector_url
+        self.agent.tracing_collector_url = self.jaeger_service.url
```
3asm, line 243: This is incorrect, the url is given in the lite local runtime.

3asm (changes requested):

PR #335, opened 2022-08-17
Add flag to expose RabbitMQ ports.

deadly-panda:


src/ostorlab/runtimes/local/runtime.py
```diff
@@ -89,14 +89,15 @@ class LocalRuntime(runtime.Runtime):
...
-        del args, kwargs
+        del args
         self.follow = []
         self._tracing = tracing
         self._mq_service: Optional[mq.LocalRabbitMQ] = None
         self._redis_service: Optional[redis.LocalRedis] = None
         self._jaeger_service: Optional[jaeger.LocalJaeger] = None
         self._log_streamer = log_streamer.LogStream()
         self._scan_db: Optional[models.Scan] = None
+        self._mq_exposed_ports: Optional[str] = kwargs.get('bind_mq_port')
```
3asm, line 100: It is better to make explicit in the args, just like tracing arg.

3asm (changes requested):

src/ostorlab/cli/scan/scan.py
```diff
@@ -66,7 +71,8 @@ def scan(ctx: click.core.Context, runtime: str,
                                                    network=network,
                                                    redis_url=redis_url,
                                                    tracing=tracing,
-                                                   tracing_collector_url=tracing_collector_url
+                                                   tracing_collector_url=tracing_collector_url,
+                                                   mq_exposed_ports = mq_exposed_ports
```
3asm, line 75: I think this should be parsed outside.

3asm (changes requested):

3asm (approved):

PR #343, opened 2022-08-27
Add fixture add_ip_network for agent testing

amine3:
Implement the add_ip_network fixture to use in the Agents unit tests.
The function has the same definition and use a Dict instead of calling Redis


3asm:
I don't think this is needed, because the new API is used already mocked APIs.

RabsonJ (approved):

PR #346, opened 2022-08-30
Fix log status

oussamaessaji:
Fix KeyError: 'status' when running install agent:
this error appear when the status it's not present in log

### DIFF
--- a/src/ostorlab/cli/agent/install/install_progress.py
+++ b/src/ostorlab/cli/agent/install/install_progress.py
@@ -23,7 +23,6 @@ class AgentInstallProgress(progress.Progress):
                                    '•',
                                    progress.TimeRemainingColumn())
 
-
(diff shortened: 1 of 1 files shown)


src/ostorlab/cli/agent/install/install_progress.py
```diff
@@ -23,7 +23,7 @@ def __init__(self):
                                    '•',
                                    progress.TimeRemainingColumn())
 
-    def display(self, logs_generator: Iterator[Dict]) -> None:
+    def display(self, logs_generator: Iterator[Dict], ) -> None:
```
3asm, line 26:
```suggestion
    def display(self, logs_generator: Iterator[Dict]) -> None:
```

3asm (approved):

PR #353, opened 2022-09-02
Fix: Bug in agent group json format

RabsonJ:
- Makes changes to the agent group schema to mark the `key` of each agent as required.
- Defaults the agent `args` to an empty list.
- Marks `name` and `type` as required in the agent schema.

### DIFF
--- a/src/ostorlab/agent/schema/agent_group_schema.json
+++ b/src/ostorlab/agent/schema/agent_group_schema.json
@@ -34,7 +34,8 @@
           "description": "[Optional] - Default value of the agent argument.",
           "type": ["string", "number", "boolean", "array", "object"]
         }
-      }
+      },
+      "required": ["name", "type"]
     },
     "agentPortMapping": {
       "description": "[Optional] - Array of dictionary-like objects, defining the port mapping.",
@@ -103,7 +104,8 @@
             "type": "array",
             "items": {
               "$ref": "#/CustomTypes/agentArgument"
-            }
+            },
+            "default": []
           },
           "port_mapping": {
             "description": "[Optional] - List of the agent port mapping.",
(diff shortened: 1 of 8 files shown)


src/ostorlab/agent/schema/agent_group_schema.json
```diff
@@ -34,7 +34,8 @@
           "description": "[Optional] - Default value of the agent argument.",
           "type": ["string", "number", "boolean", "array", "object"]
         }
-      }
+      },
+      "required": ["name", "type"]
```
3asm, line 38: And value?

src/ostorlab/agent/schema/agent_schema.json
```diff
@@ -170,7 +170,8 @@
             "description": "[Optional] - Default value of the agent argument.",
             "type": ["string", "number", "boolean", "array", "object"]
           }
-        }
+        },
+        "required": ["name", "type"]
```
3asm, line 174: This one is correct, it doesn't have value.

tests/agent/schema/test_validator.py
```diff
@@ -41,6 +46,92 @@ def testYamlValidation_whenItIsCorrect_noRaise(json_schema_file):
...
+          description: "Target that doesn't specify a schema will use this argument."
+          value: https
+    """
+    validator_object = validator.Validator(json_schema_file)
+    yaml_data_file = io.StringIO(invalid_yaml_data)
+
+    with pytest.raises(validator.ValidationError) as exc_info:
+        validator_object.validate(yaml_data_file)
+
+    assert exc_info.type is validator.ValidationError
```
3asm, line 84: Should be in the with clause.

3asm (changes requested):

tests/agent/schema/test_validator.py
RabsonJ: I'm not clear on what you mean

3asm (approved):

PR #354, opened 2022-09-02
Add pytest coverage.

3asm:
Add coverage computing and showing in the PR comment.


deadly-panda (approved):
I'm not really sure about the difference between referencing the py.typed in setup.cfg vs in setup.py.. but I'm approving for now! LGTM.

PR #356, opened 2022-09-06
fix: serializer take first value when unexpected list is given

SmileOstorlab:


src/ostorlab/agent/message/serializer.py
```diff
@@ -118,7 +118,7 @@ def _parse_list(values: Any, message: Any) -> None:
         for v in values:
             cmd = message.add()
             _parse_dict(v, cmd)
-    else:  # value can be set
+    elif isinstance(message, list):  # value can be set
```
3asm, line 121: I am not sure this code is fixing a real bug? What is the usecase you are trying to fix?

3asm (changes requested):

PR #361, opened 2022-09-16
Ensure Dict minifier returns json encodable dict.

deadly-panda:


tests/utils/dictionary_minifier_test.py
```diff
@@ -44,7 +46,7 @@ def testMinifyDict_whenNestedDict_shouldMinifyStringValues():
             },
             'key5': 5
         },
-        'key6': b'a ',
+        'key6': "b'", # pylint: disable=C4001
```
3asm, line 49: Please respect the linter, you can do 'b\''

3asm (changes requested):

tests/utils/dictionary_minifier_test.py
deadly-panda: Which does not equal the one produced by the minifiyer..

3asm (approved):

PR #376, opened 2022-10-26
Upgrade protos and add new protos

3asm:
Upgrade protos to the latest compiler version (3.21.7).
Add `v3.use.device` and `v3.capture.filesystem` protos.


src/ostorlab/agent/message/proto/v3/use/device/device.proto
```diff
@@ -0,0 +1,42 @@
...
+}
+
+message Duration {
+  optional int32 nanos = 1;
+  optional int64 seconds = 2;
+}
+
+message Message {
+  optional string device_id = 1;
+  optional string device_type = 2;
```
mohsinenar, line 36: What is device_type?

amine3: Android Iphone Ipad

amine3 (approved):
LGTM

src/ostorlab/agent/message/proto/v3/use/device/device.proto
mohsinenar: Is it the same as the platform: `android | ios`
and can it be an enum? instead of a string.

src/ostorlab/agent/message/proto/v3/use/device/device.proto
```diff
@@ -0,0 +1,42 @@
...
+  optional string address = 2;
+  optional sint32 port = 3;
+}
+
+message Duration {
+  optional int32 nanos = 1;
+  optional int64 seconds = 2;
+}
+
+message Message {
```
mohsinenar, line 34: The message is supposed to contain multiple devices, no a single device

3asm: Each device will have his own message.

PR #406, opened 2022-12-23
Add control message.

3asm:
This is a BREAKING change that requires rebuilding all the agents, but NO code change.

This adds a control message before serializing messages.

The goal is to augment the message with a list of agents in order to detect infinite loops and control message processing depth.


src/ostorlab/testing/agent.py
```diff
@@ -205,8 +209,17 @@ def agent_run_mock(mocker) -> AgentRunInstance:
 
     def mq_send_message(key, message):
         # we need to remove the last part of the key f'{selector}.{uuid.uuid1()}'
```
najibraihan, line 211: docstring

najibraihan (approved):

amine3 (approved):

src/ostorlab/testing/agent.py
3asm: This is overriding an existing documented method.

PR #415, opened 2023-01-09
Add proto definitions for android APKs details.

deadly-panda:


src/ostorlab/agent/message/proto/v3/fingerprint/file/android/activity/activity.proto
```diff
@@ -0,0 +1,23 @@
...
+  optional string permission = 2;
+  optional string process = 3;
+  optional bool enabled = 4;
+  optional bool exported = 5;
+  optional bool direct_boot_aware = 6;
+  repeated IntentFilter intent_filters= 7;
+}
+
+message Message {
+  repeated Activity activities = 1;
```
3asm, line 22: I don't think these should be separate, as they refer to a single app, and splitting them, will affect usages where you need all or some of them.

I think a single android proto that lists all of these are better. And in there have the package name for android.

Basically the proto should describe an android application.

3asm (changes requested):

3asm (approved):

PR #433, opened 2023-06-14
Add application content to use device proto.

amine3:
Add application content to use device proto.


src/ostorlab/agent/message/proto/v3/use/device/device.proto
```diff
@@ -55,4 +55,6 @@ message Message {
   optional Duration duration = 8;
   // Set of args to be passed to the device. Example are passing target to run for instance.
   repeated Argument args = 9;
+  // Application content
+  optional bytes content = 10;
```
3asm, line 59: Can you please add it to the argument? content is not device-specific. We should add support for bytes in the `Argument` instead.

3asm (changes requested):

PR #442, opened 2023-07-11
GCP remote logging.

3asm:
GCP remote logging is enabled by setting `GOOGLE_APPLICATION_CREDENTIALS` env variable.

Once the variable is set in the CLI, the runtime will pass it to the agent. Agents will then detect the presence of the variable to enable remote logging.


src/ostorlab/cli/rootcli.py
```diff
@@ -21,13 +22,20 @@
 )
 @click.option("-d", "--debug/--no-debug", help="Enable debug mode", default=False)
 @click.option("-v", "--verbose/--no-verbose", help="Enable verbose mode", default=False)
+@click.option(
+    "--gcp-logging-credential",
```
deadly-panda, line 26: Is there a reason why this is specific to GCP? Can we make the option generic, so we can support other cloud providers in the future?

deadly-panda (changes requested):

src/ostorlab/cli/rootcli.py
```diff
@@ -49,3 +57,16 @@ def rootcli(
...
+
+        with open(gcp_logging_credential, 'r') as source:
+            info = json.load(source)
+
+        credentials = service_account.Credentials.from_service_account_info(info)
+
+        # Instantiates a client
+        client = google.cloud.logging.Client(credentials=credentials)
+        client.setup_logging()
+        ctx.obj["gcp_logging_credential"] = gcp_logging_credential
```
deadly-panda, line 72:
```suggestion
        ctx.obj["gcp_logging_credential"] = gcp_logging_credential

```

src/ostorlab/agent/agent.py
```diff
@@ -18,6 +18,8 @@
 import uuid
 import json
 from typing import Dict, Any, Optional, Type, List
+import google.cloud.logging
+from google.oauth2 import service_account
```
deadly-panda, line 22: It should follow the imports convention.
```suggestion
from typing import Dict, Any, Optional, Type, List

import google.cloud.logging
from google.oauth2 import service_account
```

src/ostorlab/agent/agent.py
```diff
@@ -40,6 +44,16 @@ class MaximumCyclicProcessReachedError(exceptions.OstorlabError):
     """The cyclic process limit is enforced and reach set value."""
 
 
+def _setup_remote_logging():
+    gcp_logging_credential = os.environ.get(GCP_LOGGING_CREDENTIAL_ENV)
+    if gcp_logging_credential is not None:
+        info = json.loads(gcp_logging_credential)
+        credentials = service_account.Credentials.from_service_account_info(info)
+        # Instantiates a client
```
deadly-panda, line 52: Can we please handle the case were `json.loads` fails?
Also the comment: `# Instantiates a client` is not really useful.

src/ostorlab/cli/rootcli.py
```diff
@@ -21,13 +25,20 @@
...
+    required=False,
+)
 def rootcli(
     ctx: click.core.Context,
     proxy: Optional[str] = None,
     tlsverify: Optional[bool] = True,
     debug: bool = False,
     verbose: bool = False,
     api_key: str = None,
+    gcp_logging_credential: str = None,
```
deadly-panda, line 41:
```suggestion
    gcp_logging_credential: Optional[str] = None,
```

src/ostorlab/cli/rootcli.py
```diff
@@ -49,3 +60,13 @@ def rootcli(
         loggers = [logging.getLogger(name) for name in logging.root.manager.loggerDict]
         for l in loggers:
             l.setLevel(logging.DEBUG)
+    if gcp_logging_credential is not None:
+        with open(gcp_logging_credential, "r", encoding="utf-8") as source:
+            info = json.load(source)
+
+        credentials = service_account.Credentials.from_service_account_info(info)
+
+        # Instantiates a client
```
deadly-panda, line 69: Same as the previous comments.

deadly-panda (dismissed):

amine3 (approved):

PR #443, opened 2023-07-12
Query scanner config API

oussamaessaji:
This PR need to be updated once the API is ready

### DIFF
--- a/src/ostorlab/apis/scanner_config.py
+++ b/src/ostorlab/apis/scanner_config.py
@@ -9,8 +9,8 @@ from ostorlab.apis import request
 class GetScannerConfigAPIRequest(request.APIRequest):
     """Get scanner config."""
 
-    def __init__(self, scan_id: int):
-        self._scan_id = scan_id
+    def __init__(self, scanner_id: int):
+        self._scanner_id = scanner_id
 
     @property
     def query(self) -> Optional[str]:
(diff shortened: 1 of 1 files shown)


3asm (approved):

src/ostorlab/apis/scanner_config.py
```diff
@@ -0,0 +1,56 @@
...
+
+    @property
+    def query(self) -> Optional[str]:
+        """Defines the query to get the configs.
+
+        Returns:
+            The query to get the configs.
+        """
+        return """
+         query scanners($scannerId: Int){
```
deadly-panda, line 23: Is the scannerId parameter required ? In that case, can you please add the `!` mark at the end.

deadly-panda (changes requested):

3asm (dismissed):

src/ostorlab/apis/scanner_config.py
```diff
@@ -0,0 +1,59 @@
+"""Receive scanner config (Nats creds, Harbor creds ...)."""
+
+from typing import Dict, Optional
+import json
+
+from ostorlab.apis import request
+
+
+class GetScannerConfigAPIRequest(request.APIRequest):
```
mohsinenar, line 9: Can you find a better name? `get` is an action!

mohsinenar (changes requested):

mohsinenar (approved):

3asm (approved):

PR #447, opened 2023-07-14
Add option to disable running default agents.

deadly-panda:


src/ostorlab/runtimes/local/runtime.py
```diff
@@ -92,7 +92,7 @@ def __init__(
         *args,
         tracing: Optional[bool] = False,
         mq_exposed_ports: Optional[Dict[int, int]] = None,
-        gcp_logging_credential: Optional[str] = None,
+        run_default_agents: bool = True,
```
3asm, line 95: Unit test?

3asm (changes requested):

tests/runtimes/local/runtime_test.py
```diff
@@ -182,3 +186,43 @@ def docker_services():
...
+    assert start_agent_mock_call_args[0][0][0].key == "agent/ostorlab/agent42"
+    assert start_agent_mock_call_args[1][0][0].key == "agent/ostorlab/inject_asset"
+    assert all(
+        call_arg[0][0].key != "agent/ostorlab/local_persist_vulnz"
+        for call_arg in start_agent_mock_call_args
+    )
+    assert all(
+        call_arg[0][0].key != "agent/ostorlab/tracker"
+        for call_arg in start_agent_mock_call_args
+    )
```
3asm, line 228: Explicit checks.

3asm: I don't get these checks. Should not you rather check that there is no agent tracker in the list of agents?

3asm (changes requested):

tests/runtimes/local/runtime_test.py
deadly-panda: That's exactly what I am doing, I'm looping over all and making sure they are all `!=` tracker and persist vulnz.

3asm, line 228:
```suggestion
    assert all(
        call_arg[0][0].key != "agent/ostorlab/local_persist_vulnz"
        for call_arg in start_agent_mock_call_args
    ) is True
    assert all(
        call_arg[0][0].key != "agent/ostorlab/tracker"
        for call_arg in start_agent_mock_call_args
    ) is True
```

3asm (dismissed):

3asm (approved):

PR #448, opened 2023-07-14
Add Protobuf message definition for starting agent scans.

deadly-panda:
The compiled messages will be pushed to this PR, once the structure of the messages is agreed on!


src/ostorlab/scanner/proto/scan/_location/startAgentScan.proto
```diff
@@ -0,0 +1,64 @@
...
+import "agent/message/proto/v3/asset/file/file.proto";
+import "agent/message/proto/v3/asset/file/android/aab/aab.proto";
+import "agent/message/proto/v3/asset/file/android/apk/apk.proto";
+import "agent/message/proto/v3/asset/file/ios/ipa/ipa.proto";
+import "agent/message/proto/v3/asset/file/source/source.proto";
+import "agent/message/proto/v3/asset/agent/agent.proto";
+import "agent/message/proto/v3/asset/network/network.proto";
+import "agent/message/proto/v3/asset/link/link.proto";
+import "agent/message/proto/v3/asset/store/android_store/android_store.proto";
+import "agent/message/proto/v3/asset/store/ios_store/ios_store.proto";
```
3asm, line 16: I would suggest keeping these protos different and copying them for the moment. These can move separately and are not supposed to be the same.

3asm (changes requested):

3asm (approved):

PR #451, opened 2023-07-15
add scanner state unittest

oussamaessaji:


tests/apis/add_scanner_state_test.py
```diff
@@ -0,0 +1,31 @@
...
+    assert scanner_state_api.data["scannerState"]["scanId"] == 1
+    assert scanner_state_api.data["scannerState"]["cpuLoad"] == 50.9
+    assert scanner_state_api.data["scannerState"]["memoryLoad"] == 40.2
+    assert scanner_state_api.data["scannerState"]["totalCpu"] == 3
+    assert scanner_state_api.data["scannerState"]["totalMemory"] == 100
+    assert scanner_state_api.data["scannerState"]["hostname"] == "test"
+    assert scanner_state_api.data["scannerState"]["ipAddress"] == "0.0.0.0"
+    assert scanner_state_api.data["scannerState"]["errors"] == "error"
+
+    assert (
```
3asm, line 18: These checks are mostly checking for constants being returned. There is little value in having those, if not even bad to have them because they increase complexity of change.
A better test would integration tests that interacts with the platform.

3asm (changes requested):

PR #463, opened 2023-07-18
Call add state to persist scanner state

amine3:
Call add state to persist scanner state


src/ostorlab/cli/scanner/scanner.py
```diff
@@ -2,17 +2,39 @@
...
 
 logger = logging.getLogger(__name__)
 
+WAIT_CAPTURE_INTERVAL = 300
+
+
+async def _start_periodic_persist_state(
+    scanner_id: int, scan_id: Optional[int], errors: Optional[str]
+):
+    while True:
```
3asm, line 26: Isn't this the run method in the ScannerStateReporter?

src/ostorlab/cli/scanner/scanner.py
```diff
@@ -2,17 +2,39 @@
...
+        state_report = scanner_state_reporter.ScannerStateReporter(
+            scanner_id=int(scanner_id),
+            scan_id=scan_id,
+            hostname=socket.gethostname(),
+            ip=ip.get_public_ip(),
+            errors=errors,
```
3asm, line 32: Errors and scan_id should be read from some state class. It should not be in the constructor of the reporter.

src/ostorlab/cli/scanner/scanner.py
```diff
@@ -49,6 +78,13 @@ def scanner(
                 loop.close()
     if daemon is False and ctx.obj.get("api_key") is not None:
         loop = asyncio.get_event_loop()
+        loop.create_task(
+            _start_periodic_persist_state(
+                scanner_id=scanner_id,
+                scan_id=ctx.obj.get("scan_id"),
+                errors=ctx.obj.get("errors"),
```
3asm, line 85: Errors should not come from context.

src/ostorlab/utils/ip.py
```diff
@@ -0,0 +1,18 @@
+"""Utils to handle IP operations."""
+from typing import Optional
+
+import requests
+
+
+def get_public_ip() -> Optional[str]:
```
3asm, line 7: We must get the IP from the machine, not the public now. Absolutely not make any API call to 3rd party.

3asm (changes requested):

src/ostorlab/cli/scanner/scanner.py
```diff
@@ -49,6 +68,7 @@ def scanner(
                 loop.close()
     if daemon is False and ctx.obj.get("api_key") is not None:
         loop = asyncio.get_event_loop()
+        loop.create_task(_start_periodic_persist_state(state_report=state_report))
```
3asm, line 71: state reporter needs to get passed to subscribe_nats. Also please remove the _to_ from that method name.

3asm (changes requested):

PR #470, opened 2023-08-10
Add registry credentials to scanner conf.

deadly-panda:
The module is also renamed to a more accurate name.


src/ostorlab/scanner/scanner_conf.py
```diff
@@ -4,6 +4,14 @@
 from typing import List
 
 
+@dataclasses.dataclass
+class RegistryConfig:
+    """Represents the configuration for container registry credentials."""
+
+    username: str
+    token: str
+
```
3asm, line 13: Should also have the registry URL.

src/ostorlab/scanner/scanner_conf.py
```diff
@@ -54,9 +64,15 @@ def from_json(cls, config):
             .get("scanners", [])[0]
             .get("config", {})
         )
+        registry_conf = RegistryConfig(
+            username=conf.get("harborAccountName"),
```
3asm, line 68: Should not use harbord, please refactor to registry.

3asm (changes requested):

3asm (approved):

PR #473, opened 2023-08-11
Add login to private registry, & run the start agent callback in an async context.

deadly-panda:


src/ostorlab/scanner/callbacks.py
```diff
@@ -153,31 +165,73 @@ def _update_state_reporter(
...
-def start_scan(
+def _connect_to_containers_registry(
+    configuration: scanner_conf.RegistryConfig,
+) -> docker.DockerClient:
+    """Connect to container registry."""
+    client = docker.from_env()
+    client.login(
+        username=configuration.username,
+        password=configuration.token,
+        registry=OSTORLAB_REGISTRY,
```
3asm, line 176: :) this should be coming from the config, not hardcoded here.

3asm: Worth doing in another iteration.

src/ostorlab/scanner/callbacks.py
```diff
@@ -153,31 +165,73 @@ def _update_state_reporter(
     return state_reporter
 
 
-def start_scan(
+def _connect_to_containers_registry(
```
3asm, line 168:
```suggestion
def _connect_containers_registry(
```

3asm: Please stop using small words in function names.

3asm (changes requested):

src/ostorlab/scanner/callbacks.py
deadly-panda: It's a small change, already implemented.

3asm (approved):

PR #488, opened 2023-09-07
Add flushing of logger at the end of processing each message.

3asm:


src/ostorlab/agent/agent.py
```diff
@@ -213,6 +213,9 @@ def process_message(self, selector: str, message: bytes) -> None:
         finally:
             self.process_cleanup()
             logger.debug("done call to process message")
+            # Flush all logging handlers to ensure remote logging is sent before app shutdown.
+            for h in logger.handlers:
```
amine3, line 217: missing unittest to check it is called

amine3 (changes requested):

src/ostorlab/agent/agent.py
3asm: How do you suggest I test this?

amine3 (approved):

PR #494, opened 2023-09-25
Feature/Report CVE  protobuf message definition.

deadly-panda:


src/ostorlab/agent/message/proto/v3/report/cve/cve.proto
```diff
@@ -0,0 +1,51 @@
+syntax = "proto2";
+
+package ostorlab.agent.message.proto.v3.report.cve;
+
+message Message {
+    optional int32 scan_id = 1;
```
3asm, line 6: Why scan_id?

src/ostorlab/agent/message/proto/v3/report/cve/cve.proto
```diff
@@ -0,0 +1,51 @@
...
+    optional string cve_id = 2;
+    repeated Target targets = 3;
+    optional int32 cwe = 4;
+    optional string description = 5;
+    repeated Reference references = 6;
+    optional string cvss_v3_1_vector = 7;
+    optional string cvss_v3_0_vector = 8;
+    optional string cvss_v2_vector = 9;
```
3asm, line 14: Should be grouped into a cvss message.

src/ostorlab/agent/message/proto/v3/report/cve/cve.proto
```diff
@@ -0,0 +1,51 @@
...
+    optional string published_date = 10;
+    optional string modified_date = 11;
```
3asm, line 16: int into epoch is better.

3asm (changes requested):

src/ostorlab/agent/message/proto/v3/report/cve/cve.proto
3asm: Also the cvss should be broken down.

src/ostorlab/agent/message/proto/v3/report/cve/cve.proto
```diff
@@ -0,0 +1,105 @@
...
+
+message Message {
+    optional string cve_id = 1;
+    repeated Target targets = 2;
+    optional int32 cwe = 3;
+    optional string description = 4;
+    repeated Reference references = 5;
+    optional CvssScoreV3 cvss_v3_1_vector = 6;
+    optional CvssScoreV3 cvss_v3_0_vector = 7;
+    optional CvssScoreV2 cvss_v2_vector = 8;
```
3asm, line 13: It would be nicer to group them and have a naming convention like cvss.v3_0. No need for vector in the name.

3asm (changes requested):

src/ostorlab/agent/message/proto/v3/report/cve/cve.proto
```diff
@@ -0,0 +1,111 @@
...
+    optional CvssScores cvss_scores = 6;
+    optional uint64 published_date = 7;
+    optional uint64 modified_date = 8;
+}
+
+message CvssScores { 
+    optional CvssScoreV3 cvss_v3_1 = 1;
+    optional CvssScoreV3 cvss_v3_0 = 2;
+    optional CvssScoreV2 cvss_v2 = 3;
+}
```
3asm, line 20:
```suggestion
message CVSS { 
    optional CVSSv31 v31 = 1;
    optional CVSSv30 v30 = 2;
    optional CVSSv2 v2 = 3;
}
```

3asm (changes requested):

src/ostorlab/agent/message/proto/v3/report/cve/cve.proto
benyissa: Do we really need both CVSSV3.1 and CVSSV3.0 versions or keep only the latest one since we need to persist with only one?
And what if other versions occur? CVSS v4.0, for example, that could be found here : https://www.first.org/cvss/v4-0/index.html

3asm: We need all.

3asm (approved):

ErebusZ (approved):

PR #513, opened 2023-10-30
Handle same agent added multiple times in the agent group definition.

amine3:
We add part of the uuid to avoid conflict when having the same agent added multiple times in the agent group definition

### DIFF
--- a/src/ostorlab/runtimes/lite_local/agent_runtime.py
+++ b/src/ostorlab/runtimes/lite_local/agent_runtime.py
@@ -349,7 +349,9 @@ class AgentRuntime:
             + "_"
             + self.runtime_name
             + "_"
-            + str(uuid.uuid4())[:7]
+            + str(uuid.uuid4())[
+                :7
+            ]  # We add only the first characters due the 64 characters limitations.
         )
 
         env = [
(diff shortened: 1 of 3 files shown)


elyousfi5 (dismissed):
lgtm

src/ostorlab/runtimes/lite_local/agent_runtime.py
```diff
@@ -348,6 +348,10 @@ def create_agent_service(
             or self.agent.container_image.replace(":", "_").replace(".", "")
             + "_"
             + self.runtime_name
+            + "_"
+            + str(uuid.uuid4())[
```
3asm, line 352: This will cause problem, there is a limit on the service name size. The value should be incremental with .1 .2. 3.

3asm (changes requested):

src/ostorlab/runtimes/lite_local/agent_runtime.py
3asm: Also, this updating on the lite_local runtime, it should be applied to all runtimes!

src/ostorlab/runtimes/lite_local/agent_runtime.py
```diff
@@ -348,6 +349,8 @@ def create_agent_service(
             or self.agent.container_image.replace(":", "_").replace(".", "")
             + "_"
             + self.runtime_name
+            + "_"
+            + "".join([str(random.randint(1, 9)) for _ in range(5)])
```
3asm, line 353:
```suggestion
            + str(random.randrange(0, 9999))
```

3asm (changes requested):

3asm (approved):

PR #516, opened 2023-10-31
Fix/change the behavior to fail if an argument is present in the settings but not in the agent definition 

oussamaessaji:


src/ostorlab/agent/agent.py
```diff
@@ -156,6 +160,7 @@ def run(self) -> None:
 
         Connects to the agent bus, start health check and start listening to new messages.
         """
+        breakpoint()
```
rabsondev, line 163: Not needed, right?

rabsondev (changes requested):

src/ostorlab/agent/agent.py
oussamaessaji: exactly

3asm:
We can't do this yet.

src/ostorlab/agent/agent.py
```diff
@@ -47,6 +47,10 @@ class MaximumCyclicProcessReachedError(exceptions.OstorlabError):
     """The cyclic process limit is enforced and reach set value."""
 
 
+class ArgumentMissingInAgentDefinitionError(exceptions.OstorlabError):
+    """Argument is present in the settings but not in the agent definition."""
```
deadly-panda, line 51: Can be put in the exceptions module, since it's used in multiple places.

tests/runtimes/local/agent_runtime_test.py
```diff
@@ -149,3 +151,66 @@ def testCreateAgentService_whenAgentDefIsNotEmptyAndAgentSettingsIsEmpty_service
     assert kwargs["mounts"] == ["def_mount1", "def_mount2"]
     assert kwargs["endpoint_spec"]["Ports"][0]["PublishedPort"] == 30000
     assert kwargs["restart_policy"]["Condition"] == "any"
+
+
+def testLoaclRunTime_whenAgentArgsAreWrong_shouldFail(
```
ErebusZ, line 156: where is the test for lite local runtime?

ErebusZ (changes requested):

src/ostorlab/agent/agent.py
```diff
@@ -129,7 +129,6 @@ def args(self) -> Dict[str, Any]:
         for a in self.settings.args:
             # Enforce that only declared arguments are accepted.
             if a.name not in arguments:
-                # TODO(OS-5119): Change behavior to fail of the argument is missing from definition.
                 logger.warning(
```
amine3, line 132: Why do we not raise here the error?

amine3 (changes requested):
We need to fix inject store agent before merging this one

amine3 (approved):

rabsondev (approved):
lgtm

ErebusZ (approved):

PR #517, opened 2023-10-31
Catch CONNECTION_EXCEPTIONS with  MQ

benyissa:
testing use case : 

step1 : rebuild nvd agent and tracker in local using current branch fix_mq_connection_errors : 

```
RUN apk add git
RUN git clone -b fix_mq_connection_errors https://github.com/Ostorlab/ostorlab
RUN pip install -e /ostorlab[agent]
```
step 2 : lunch a scan with the following command : 

`ostorlab scan run  --no-asset --agent agent//nvd_collector --agent agent//nvd_persist --agent agent//tracker --follow agent//nvd_collector --follow agent//nvd_persist --follow agent//tracker`

image

image

messages are in MQ now  and nvd persist is processing them .

step3 : 

stop MQ container using `docker stop container-id`

step 4 : 
agents are not crashing but try to reconnect instead : 

image

side errors : 

Error 1 : 

Tracker agent try to connect with MQ using HTTP here : https://github.com/Ostorlab/agent_tracker/blob/main/agent/request_sender.py
image

==> we need to increase retry or do infinity loop as we did with the current fix .
Error 2: 
when the connection reestablished tracker will kill the universe directly since there is no message in MQ (new instance created if the error is really MQ instance is down)
==> we need to increase MAX_COUNT and SLEEP_SEC here :  https://github.com/Ostorlab/agent_tracker/blob/main/agent/data_queues.pyhttps://github.com/Ostorlab/agent_tracker/blob/main/agent/data_queues.py

Error 3 :
when connection is already established and we try to publish or proccess a message . the current async task will be canceled and we could have errors like this one : ChannelInvalidStateError
for more detailsl please check this issue in aio pika : https://github.com/mosquito/aio-pika/issues/508
image

==> we need to catch them too and recall the current function.


src/ostorlab/agent/mixins/agent_mq_mixin.py
```diff
@@ -6,12 +6,17 @@
...
+import datetime
 from typing import List, Optional
 
 import aio_pika
+import tenacity
 
 logger = logging.getLogger(__name__)
 
+MAX_MQ_CALL_RETRIES = 5
+MQ_CALL_WAIT_BEFORE_RETRY = datetime.timedelta(seconds=120)
```
ErebusZ, line 18: why use datetime?

elyousfi5 (dismissed):

3asm:
Awesome work @benyissa .

src/ostorlab/agent/mixins/agent_mq_mixin.py
```diff
@@ -50,15 +51,18 @@ def __init__(
             self._get_channel, max_size=64, loop=self._loop
         )
 
-    async def _get_connection(self) -> aio_pika.Connection:
-        return await aio_pika.connect_robust(url=self._url, loop=self._loop)
+    async def _get_connection(self) -> aio_pika.abc.AbstractRobustConnection:
+        kwargs = {"fail_fast": "false"}
+        return await aio_pika.connect_robust(url=self._url, loop=self._loop, **kwargs)
```
3asm, line 56:
```suggestion
        return await aio_pika.connect_robust(url=self._url, loop=self._loop, fail_fast=False)
```

Why is false a string and not a bool?

3asm (changes requested):

src/ostorlab/agent/mixins/agent_mq_mixin.py
benyissa: They didn't support flexible parsers before, but this has been fixed in the current version.
https://github.com/mosquito/aiormq/commit/1cdfb06d58d6be91028207b62d9a47653da4d199

Good catch. I will update it to false instead.

3asm (approved):

PR #521, opened 2023-11-02
Apply random value to service name to handle the same agent declared multiple times.

amine3:


tests/runtimes/local/agent_runtime_test.py
```diff
@@ -83,6 +83,7 @@ def testCreateAgentService_whenAgentDefAndAgentSettingsAreNotEmpty_serviceCreate
 
     # assert arguments were overridden by the agent settings.
     assert kwargs["resources"]["Limits"]["MemoryBytes"] == 700000
+    assert len(kwargs["name"]) < 63
```
3asm, line 86: Can you assert on the name value. We need to ensure we don't end up with the same name.

3asm (changes requested):

3asm (approved):

PR #540, opened 2023-12-07
Fix the creation of agent group from list.

3asm:


src/ostorlab/runtimes/cloud/runtime.py
```diff
@@ -492,7 +492,7 @@ def _check_agents_exist(
         return True
 
     def _to_serialized(self, value) -> Union[bytes, memoryview, List[Any], str]:
-        has_bytes_scalar_type = isinstance(value, (bytes, list, memoryview, str))
+        has_bytes_scalar_type = isinstance(value, (bytes, memoryview))
```
elyousfi5, line 495:
```suggestion
        has_bytes_scalar_type = isinstance(value, (bytes, memoryview)) is True
```

PR #553, opened 2024-01-10
Use protobuf to dict instead of implicit access + other minor fixes

elyousfi5:


3asm (approved):
Nice work and good testing coverage.

PR #554, opened 2024-01-16
Catch ConnectionResetError.

mohsinenar:


benyissa (dismissed):
The error message `ERROR Task exception was never retrieved` typically occurs when an exception is raised within an asynchronous task, but the exception is not explicitly caught or awaited, causing it to be missed.

currently in the code base we call `async_mq_send_message` in 2 cases  : 

if the event loop is currently running we use  `create_task` this will never raise the exception but it will be missed as explained above. 

if the event loop is currently running we use  `run_until_complete` this will return the Future's result, or raise its exception , in this case tenacity can be helpful.

if we want to catch them we can catch ConnectionError directly. but this will never fix this kind of issues if the `async_mq_send_message` was called using `create_task`.

src/ostorlab/agent/mixins/agent_mq_mixin.py
```diff
@@ -168,15 +168,17 @@ async def async_mq_send_message(
             await exchange.publish(routing_key=key, message=pika_message)
 
     @tenacity.retry(
-        retry=tenacity.retry_if_exception_type(aio_pika.exceptions.ConnectionClosed),
+        retry=tenacity.retry_if_exception_type(
```
3asm, line 171: Needs unit test.

3asm (changes requested):

benyissa (commented):
A new release for both `aiormq` and `aio-pika` was published 2 days ago. You can find the releases on the following GitHub repositories:

- `aiormq`: [https://github.com/mosquito/aiormq](https://github.com/mosquito/aiormq)
- `aio-pika`: [https://github.com/mosquito/aio-pika](https://github.com/mosquito/aio-pika)

We should review the change logs to determine if there are any modifications that could impact our current logic. If any such changes exist, it's advisable to pin the version accordingly.

3asm:
Minor updates, just python version changes.

tests/agent/mixins/agent_mq_mixin_test.py
```diff
@@ -126,3 +126,19 @@ async def testClient_whenClientDisconnects_messageIsNotLost(mocker, mq_service):
     # make sure the message is received and was not deleted
     stub.assert_called_with(word)
     assert stub.call_count == 1
+
+
+def testMqSendMessage_onConnectionResetError_shouldRetriesAndReraise(mocker):
```
elyousfi5, line 131: typing please

3asm (approved):

PR #557, opened 2024-01-23
Feat: Use Ruff as a linter and formatter

benyissa:
In this pull request, the decision was made to exclusively transition from pylint and black to utilizing Ruff, a high-speed Python linter and code formatter written in Rust, due to its rapid feedback loop, comprehensive rule set, and compatibility with popular code editors, aiming to accelerate checks and improve the development flow.

### DIFF
--- a/src/ostorlab/cli/ci_scan/__init__.py
+++ b/src/ostorlab/cli/ci_scan/__init__.py
@@ -2,4 +2,4 @@
 from ostorlab.cli.ci_scan import run
 from ostorlab.cli.ci_scan.ci_scan import ci_scan
 
-__all__ = [run, ci_scan()]
+__all__ = [run, ci_scan]


src/ostorlab/cli/ci_scan/__init__.py
```diff
@@ -1,3 +1,5 @@
 """Module for the root command ci_scan"""
 from ostorlab.cli.ci_scan import run
 from ostorlab.cli.ci_scan.ci_scan import ci_scan
+
+__all__ = [run, ci_scan()]
```
oussamaessaji, line 5:
```suggestion
__all__ = [run, ci_scan]
```

benyissa: allright! That's why pytest wasn't running correctly.

src/ostorlab/cli/rootcli.py
```diff
@@ -54,12 +54,12 @@ def rootcli(
 
     if verbose is True:
         loggers = [logging.getLogger(name) for name in logging.root.manager.loggerDict]
-        for l in loggers:
-            l.setLevel(logging.INFO)
+        for current_logger in loggers:
```
ErebusZ, line 57:
```suggestion
        for logger in loggers:
```

src/ostorlab/cli/__init__.py
```diff
@@ -6,3 +6,5 @@
 from ostorlab.cli import agentgroup
 from ostorlab.cli import ci_scan
 from ostorlab.cli import scanner
+
+__all__ = [scan, auth, agent, vulnz, agentgroup, ci_scan, scanner]
```
ErebusZ, line 10: I would favor excluding the rule from ruff instead of adding `__all__` to the modules
this exporting method only affects the behavior of `from module import *` and is not used by us in anyway. plus the potential of future breakage.

fixing the error like this would just impose unused extra code.

typing_requirements.txt
```diff
@@ -0,0 +1,8 @@
+mypy
```
ErebusZ, line 1: why add a specific file for the typing requirements?

ErebusZ (changes requested):

src/ostorlab/cli/__init__.py
3asm: Nah, this is much better.

3asm:
Nice work.

typing_requirements.txt
benyissa: it's used in the mypy workflow.
and we need them while working on local to verify mypy issues.

PR #570, opened 2024-02-15
Add Agent_settings in selectors

elyousfi5:
For testing i removed in_selectors all in_selectors of agent tsunami and left just `v3.asset.ip.v4`:
image

In agent group definition i passed `v3.asset.link`:
image

The right behavior of the agent would be to process messages of selector `v3.asset.link` and dont process messages of selector `v3.asset.ip.v4`, since they are overrided :

Link message test:
image

Ipv4 message test:
image


src/ostorlab/agent/agent.py
```diff
@@ -395,6 +395,37 @@ def _prepare_message(self, raw: bytes) -> bytes:
...
+            for selector in in_selectors
+            if any(
+                selector in extended_selector
+                for extended_selector in extended_selectors
+            )
+            is False
+        ]
+
+        in_selectors.extend(extended_selectors)
+
```
3asm, line 426: extended selectors should simply override the selectors. For this reason I don't think `extended` is the right name in the rest of the code.

The behavior we want to achieve is to replace the existing ones with the provided ones.

3asm (changes requested):

src/ostorlab/agent/agent.py
elyousfi5: @3asm i will rename to `runtime_in_selectors` !

3asm (approved):

PR #575, opened 2024-02-19
Use tmp dir to inject assets instead of temporary container.

mohsinenar:


3asm (approved):

src/ostorlab/utils/volumes.py
```diff
@@ -26,62 +18,29 @@ def __init__(self) -> None:
...
-            detach=True,
-            remove=True,
-            mounts=[
-                docker_types.Mount(
-                    target=TEMP_DESTINATION, source=volume_name, type="volume"
-                )
-            ],
+        # Create a new directory in the system's temporary directory
+        volume_path = pathlib.Path(tempfile.gettempdir()) / volume_name
+        volume_path.mkdir(parents=True, exist_ok=True)
```
oussamaessaji, line 35: If the folder already exists, a new folder needs to be generated and returned.

PR #576, opened 2024-02-20
Create agent service in replicated (no need to scale separately).

mohsinenar:
image


src/ostorlab/runtimes/lite_local/agent_runtime.py
```diff
@@ -381,6 +384,7 @@ def create_agent_service(
             endpoint_spec=endpoint_spec,
             resources=docker_types_services.Resources(mem_limit=mem_limit),
             cap_add=caps,
+            mode=docker_types_services.ServiceMode("replicated", replicas),
```
deadly-panda, line 387: Can you please try playing with the `concurrency` argument, see if has any significant impact on startup performance?

3asm (approved):

PR #584, opened 2024-02-21
Revert "Use tmp dir to inject assets instead of a temporary container. """

deadly-panda:
Reverts Ostorlab/ostorlab#583


ErebusZ (approved):

benyissa (approved):
Don't forget to mention why in the prescription, please.

3asm (approved):

PR #596, opened 2024-03-05
Fix retry timeout to check the agent status.

amine3:


deadly-panda (approved):

benyissa (commented):
Could you explain why the switch to a fixed time retry strategy is preferred in this case?

PiranhaSa (approved):

amine3:
> Could you explain why the switch to a fixed time retry strategy is preferred in this case?
With the inital config, we would do
1st attempt: 1 second
2nd attempt: 2 seconds
3rd attempt: 4 seconds
In the case of the agent becoming ready after 4s, we need to wait 7s loosing 3s. wait_exponential is interesting if we do the check multiple times and we do not want to stress the system while here it is done only once and we want to converge asap.

src/ostorlab/runtimes/local/runtime.py
```diff
@@ -434,7 +434,7 @@ def _start_agent(
 
     @tenacity.retry(
         stop=tenacity.stop_after_attempt(20),
-        wait=tenacity.wait_exponential(multiplier=1, max=12),
+        wait=tenacity.wait_fixed(0.5),
```
deadly-panda, line 437: We may need to increase the `stop_after_attempt` though..

amine3: Good point, I will check in the case of multiple agents.

PR #609, opened 2024-03-07
Start tracker with other agents.

amine3:


src/ostorlab/runtimes/local/runtime.py
```diff
@@ -204,6 +204,10 @@ def scan(
             console.info("Starting agents")
             self._start_agents(agent_group_definition)
 
+            if self._run_default_agents is True:
```
mohsinenar, line 207: How about starting all the default agents at once? can you try that?

amine3: This is what we do with this config, we start pre-agents, the agents and post-agents at the same time.

mohsinenar: @amine3 what I mean is to start pre-agents and post-agents in the same call! 
```python
 if self._run_default_agents is True:
                console.info("Starting pre-agents")
                self._start_pre_agents()
                self._start_post_agents()
```

3asm: The separation makes is semantic to differentiate between the agents' types. Once the flow is confirmed to work. We may merge API.

3asm (approved):

PR #612, opened 2024-03-08
Merge windows and linux pytest workflow

ErebusZ:


.github/workflows/pytest.yml
```diff
@@ -39,7 +39,12 @@ jobs:
...
-    - name: Running tests with pytest.
+    - name: Running tests with pytest for linux.
       run: |
         set -o pipefail
         pytest -m "not docker and not nats"
+      if: matrix.os == 'ubuntu-latest'
+    - name: Running tests with pytest for windows.
+      run: |
+        pytest -m "not docker and not nats"
+      if: matrix.os == 'windows-latest'
```
3asm, line 50: If we need to do if else, then keep the flows separate.

3asm (changes requested):

PR #629, opened 2024-03-14
Add Installers workflow

ErebusZ:


.github/workflows/mac_installer.yaml
```diff
@@ -0,0 +1,35 @@
+on:
+  release:
+    types: [published]
+
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
```
deadly-panda, line 10: Why false ?

ErebusZ: originally I had one workflow for both mac and ubuntu, this is no longer the case. I ll remove it.

benyissa (changes requested):
We don't need a matrix strategy since there is only one configuration.

.github/workflows/mac_installer.yaml
```diff
@@ -0,0 +1,34 @@
...
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies.
+      run: |
+        python -m pip install --upgrade pip
+        pip install .
+        pip install .[scanner]
+        pip install .[agent]
+
+    - run: |
```
benyissa, line 29:
```suggestion
    - name: Create DMG file
    - run: |
```

3asm: +1

3asm (approved):

3asm (approved):

PR #639, opened 2024-03-22
Only encode non-bytes values in persist mock fixture.

deadly-panda:


src/ostorlab/testing/agent.py
```diff
@@ -100,7 +100,9 @@ def _add(key, value):
         storage[key] = str(value).encode()
 
     def _hash_add(hash_name, mapping):
-        mapping = {k: str(v).encode() for k, v in mapping.items()}
+        for k, v in mapping.items():
+            if isinstance(v, bytes) is False:
+                mapping[k] = str(v).encode()
```
3asm, line 105: You forgot to set the mapping!

PR #648, opened 2024-04-03
Change Werkzeug pinning

ErebusZ:


setup.cfg
```diff
@@ -50,7 +50,7 @@ install_requires =
     docker
     importlib-metadata; python_version<"3.8"
     jsonschema>=4.4.0
-    protobuf~=3.20
+    protobuf==3.20.1
```
BlueSquare1, line 53:
```suggestion
    protobuf~=3.20
```
Why not use compatible version like `Werkzeug`

ErebusZ: what do you mean compatible version?
also that change is intended.

benyissa (commented):
what about setuptools in testing module since it's reverted too?

elyousfi5 (dismissed):

ErebusZ:
> What about setuptools in the testing module since it's been reverted too?

Setuptools was changed by changing the base image in a previous PR.

3asm (approved):

PR #669, opened 2024-05-28
Add testflight proto and asset.

amine3:


src/ostorlab/cli/scan/run/assets/ios_testflight.py
```diff
@@ -0,0 +1,27 @@
...
+logger = logging.getLogger(__name__)
+
+
+@run.run.command(name="ios-testflight")
+@click.option("--application-url", required=False)
+@click.pass_context
+def ios_store(ctx: click.core.Context, application_url: str) -> None:
+    """Run scan using testflight url."""
+    runtime = ctx.obj["runtime"]
+    assets = [ios_testflight_asset.IOSTestflight(application_url=application_url)]
```
oussamaessaji, line 21: you need to check if application_url is None since IOSTestflight accept str

amine3: It is a mandatory field so it cannot be called with a None value

3asm (approved):

PR #686, opened 2024-06-03
Fix workflows getting cancelled because of overlapping group name

deadly-panda:
Fix workflows getting canceled.
image


benyissa (commented):
Could you please explain why you deleted the Mac installer?

.github/workflows/mac_installer.yaml
mohsinenar, line 1: Why this was removed?

mohsinenar (changes requested):

.github/workflows/mac_installer.yaml
deadly-panda: The installers for mac & Linux are grouped in `release_linux_mac_bin.yaml`. this `mac_installer.yaml` does not even work. I supposed it was from the time the feature was being added.

3asm (approved):

PR #688, opened 2024-06-03
Set check_same_thread to false.

elyousfi5:
This PR fixes the following error:
```python
ProgrammingError: SQLite objects created in a thread can only be used in that same thread. The object was created in thread id 124220790916672 and this is thread id 124220869264000.
```

check_same_thread set to `True`: Ensures SQLite connections are thread-safe, restricting usage to the thread that created them.
check_same_thread set to `False`: Allows SQLite connections to be shared across threads.


benyissa (commented):
Does the error happen when you try to kill the server with `ctrl + c` or in which case?

elyousfi5:
> Does the error happen when you try to kill the server with `ctrl + c` or in which case?

Nope, as @mohsinenar its not the only case.

PiranhaSa (approved):

oussamaessaji (commented):
the connections should be thread safe!!

PR #698, opened 2024-06-06
Convert arg value to bytes

elyousfi5:


mohsinenar (approved):

src/ostorlab/runtimes/local/runtime.py
mohsinenar, line 1: You should add unit tests!

mohsinenar (changes requested):

src/ostorlab/runtimes/local/runtime.py
oussamaessaji: +1

PR #701, opened 2024-06-10
Link cli scan with asset and agent group.

elyousfi5:


src/ostorlab/runtimes/local/models/models.py
```diff
@@ -572,6 +650,44 @@ class Asset(Base):
         sqlalchemy.Integer, sqlalchemy.ForeignKey("scan.id"), nullable=True
     )
 
+    @staticmethod
+    def create_from_assets_def(
+        scan: Scan, assets: Optional[List[base_asset.Asset]]
```
benyissa, line 655: 1. I would use only scan_id  in the param.
2. can't we create an asset without scan ?

elyousfi5: Yes we can, using just create

elyousfi5: Applied

src/ostorlab/cli/scan/run/assets/android_apk.py
```diff
@@ -48,10 +50,24 @@ def android_apk(
...
     logger.debug("scanning assets %s", [str(asset) for asset in assets])
     try:
-        runtime.scan(
+        created_scan = runtime.scan(
             title=ctx.obj["title"],
             agent_group_definition=ctx.obj["agent_group_definition"],
             assets=assets,
         )
+
+        if isinstance(ctx.obj["runtime"], runtime_local.LocalRuntime) is True:
```
3asm, line 59: As discussed, update runtime to add a method to perform the action and the other runtimes don't do anything in this case.

3asm (changes requested):

src/ostorlab/runtimes/local/models/models.py
```diff
@@ -412,18 +424,51 @@ def create(
...
+    @staticmethod
+    def to_bytes(
+        type: str, value: Optional[Union[int, float, str, bool, List, Dict]]
+    ) -> bytes:
+        """Convert the value to bytes."""
+        if type == "string":
+            return value.encode(encoding="utf-8")
+        elif type in ("boolean", "bool"):
+            return (1 if value is True else 0).to_bytes(1, byteorder="big")
+        elif type in ("number", "int", "float"):
```
deadly-panda, line 451: We only support `number`

src/ostorlab/runtimes/local/models/models.py
```diff
@@ -412,18 +424,51 @@ def create(
...
+            return struct.pack("d", value)
+        elif type in ("array", "object"):
```
deadly-panda, line 453: am not sure about `object`?

src/ostorlab/runtimes/local/runtime.py
```diff
@@ -769,3 +768,34 @@ def dump_vulnz(self, scan_id: int, dumper: dumpers.VulnzDumper):
...
+        with models.Database() as session:
+            agent_group_db = models.AgentGroup.create_from_agent_group_def(
+                agent_group_definition
+            )
+            scan.agent_group_id = agent_group_db.id
+            session.add(scan)
+            session.commit()
+
+    def link_assets_scan(
+        self, scan_id: int, assets: Optional[List[base_asset.Asset]]
```
deadly-panda, line 792:
```suggestion
        self, scan_id: int, assets: Optional[List[base_asset.Asset]] = None
```

src/ostorlab/runtimes/local/models/models.py
```diff
@@ -472,6 +517,39 @@ def create(name: str, description: str, agents: Any) -> "AgentGroup":
             session.commit()
             return agent_group
 
+    @staticmethod
+    def create_from_agent_group_def(
```
3asm, line 521:
```suggestion
    def create_from_agent_group_definition(
```

src/ostorlab/runtimes/local/models/models.py
```diff
@@ -572,6 +650,44 @@ class Asset(Base):
         sqlalchemy.Integer, sqlalchemy.ForeignKey("scan.id"), nullable=True
     )
 
+    @staticmethod
+    def create_from_assets_def(
```
3asm, line 654:
```suggestion
    def create_from_assets_definition(
```

src/ostorlab/runtimes/runtime.py
```diff
@@ -96,3 +96,25 @@ def dump_vulnz(self, scan_id: int, dumper: dumpers.VulnzDumper):
         None
         """
         raise NotImplementedError()
+
+    def link_agent_group_scan(
```
3asm, line 100: Should be marked as abstract method.

3asm (changes requested):

tests/conftest.py
```diff
@@ -1100,3 +1106,81 @@ def ios_store(mocker: plugin.MockerFixture, db_engine_path: str) -> models.IosSt
...
+
+    mocker.patch(
+        "ostorlab.runtimes.local.runtime.LocalRuntime.can_run", return_value=True
+    )
+    mocker.patch("ostorlab.runtimes.local.runtime.LocalRuntime._create_network")
+    mocker.patch("ostorlab.runtimes.local.runtime.LocalRuntime._start_services")
+    mocker.patch("ostorlab.runtimes.local.runtime.LocalRuntime._start_pre_agents")
+    mocker.patch("ostorlab.runtimes.local.runtime.LocalRuntime._start_agents")
+    mocker.patch("ostorlab.runtimes.local.runtime.LocalRuntime._start_post_agents")
+    mocker.patch("ostorlab.runtimes.local.runtime.LocalRuntime._check_services_healthy")
```
deadly-panda, line 1181: Can we avoid mocking private methods? or is it too much ?

src/ostorlab/runtimes/local/models/models.py
elyousfi5: in the schema yes, but in some agents we use others. Example nmap:
image

src/ostorlab/runtimes/local/models/models.py
elyousfi5: what about it ? 
image

tests/conftest.py
elyousfi5: I tried, there is no way to get over it !

src/ostorlab/runtimes/local/models/models.py
```diff
@@ -412,18 +424,51 @@ def create(
...
 
+    @staticmethod
+    def to_bytes(
+        type: str, value: Optional[Union[int, float, str, bool, List, Dict]]
+    ) -> bytes:
+        """Convert the value to bytes."""
+        if type == "string":
+            return value.encode(encoding="utf-8")
+        elif type in ("boolean", "bool"):
+            return (1 if value is True else 0).to_bytes(1, byteorder="big")
```
mohsinenar, line 450: Why not bytes(value)??

src/ostorlab/runtimes/local/models/models.py
```diff
@@ -572,6 +650,44 @@ class Asset(Base):
...
+
+        Args:
+            scan: The scan object.
+            assets: The list of assets to create.
+        """
+        networks: List[str] = []
+        links: List[str] = []
+        for asset in assets:
+            if (
+                isinstance(asset, ip.IP)
```
mohsinenar, line 667: Explicit ?

mohsinenar (dismissed):

src/ostorlab/runtimes/local/models/models.py
elyousfi5: im using this method to make it easy to get value from bytes

src/ostorlab/runtimes/local/models/models.py
deadly-panda: We do not force the type check when schema validation? how it was not spotted ..
The agent & the schema should be aligned..

3asm: +1

3asm (dismissed):

3asm (approved):

PR #715, opened 2024-06-13
Add domain asset model

elyousfi5:


src/ostorlab/serve_app/oxo.py
```diff
@@ -372,6 +373,9 @@ def mutate(
             if asset.ip is not None:
                 new_asset = models.Network.create(networks=asset.ip)
                 created_assets.append(new_asset)
+            if asset.domain is not None:
+                new_asset = models.DomainAsset.create(domains=asset.domain)
```
3asm, line 377: Why is this a list?

3asm: Why not have a single DomainAsset and actually fix that also in network and return a list of Networks?

3asm (changes requested):

tests/serve_app/oxo_test.py
```diff
@@ -2243,6 +2292,81 @@ def testRunScanMutation_whenNetworkAsset_shouldRunScan(
...
+        "ostorlab.cli.docker_requirements_checker.is_swarm_initialized",
+        return_value=True
+    )
+    mocker.patch("docker.from_env")
+    mocker.patch(
+        "ostorlab.runtimes.local.runtime.LocalRuntime.can_run", return_value=True
+    )
+    scan_mock = mocker.patch(
+        "ostorlab.runtimes.local.runtime.LocalRuntime.scan", return_value=scan
+    )
```
deadly-panda, line 2320: I think I've seen this as a fixture before, can you please double check..

elyousfi5: applied

src/ostorlab/serve_app/oxo.py
elyousfi5: Im not sure i do understand exactly your comment. 

it is a list because in the yaml file we provide it as a list, so we need to adapt  :
image

and its the same for networks. If you check unit tests we do indeed return a list of networks/domains ..

If this doesnt answer your comment please clarify.

3asm: Nothing says we should create ONE asset from that presentation.

PiranhaSa (approved):
:+1:

3asm (approved):

PR #719, opened 2024-06-14
Fix agents pagination

elyousfi5:


src/ostorlab/serve_app/types.py
```diff
@@ -751,7 +751,7 @@ def resolve_agents(
                     has_next=page.has_next(),
                     has_previous=page.has_previous(),
                 )
-                return OxoVulnerabilitiesType(agents=page, page_info=page_info)
+                return AgentsType(agents=page, page_info=page_info)
```
3asm, line 754: Should be OxoAgentsType

3asm (changes requested):

3asm (approved):

PR #723, opened 2024-06-20
Create predefined agent groups just once.

elyousfi5:


src/ostorlab/serve_app/agent_groups/autodiscovery.yaml
```diff
@@ -1,6 +1,6 @@
 kind: AgentGroup
 description: Enumerate domain scan
-name: autodiscovery
+name: predefined_oxo_autodiscovery
```
rabsondev, line 3: The `name` is required, right? The examples in https://oxo.ostorlab.co/tutorials/examples don't have `name`. So we should either update them 👍 or make `name` optional 👎

elyousfi5: as discussed this has nothing to do with optional.

src/ostorlab/runtimes/local/models/models.py
```diff
@@ -619,23 +619,34 @@ def create_from_directory(agent_groups_path: pathlib.Path) -> List["AgentGroup"]
...
+                ):
+                    with open(agent_group_file, "r") as file:
+                        agent_group_definition = (
+                            definitions.AgentGroupDefinition.from_yaml(file)
+                        )
+                        count_agent_group = (
+                            session.query(AgentGroup)
+                            .filter_by(
+                                name=agent_group_definition.name,
+                                description=agent_group_definition.description,
```
3asm, line 637: We can ignore the description.

3asm (changes requested):

3asm (approved):

PR #727, opened 2024-06-25
Resolve yaml source

elyousfi5:


src/ostorlab/serve_app/types.py
```diff
@@ -543,6 +546,60 @@ def resolve_asset_types(
...
+    ) -> str:
+        """Resolve yaml source query.
+        Args:
+            self (models.AgentGroup): The agent group object.
+            info (graphql_base.ResolveInfo): GraphQL resolve info.
+        Returns:
+            str: The yaml source of the agent group.
+        """
+        yaml = ruamel.yaml.YAML(typ="safe")
+        yaml.width = 100000000
```
rabsondev, line 560: I would put this number in a const.

elyousfi5: applied

3asm (approved):

PR #728, opened 2024-06-26
Make asset types enum

elyousfi5:


src/ostorlab/runtimes/local/models/alembic/versions/69c7d2f49a4e_switch_asset_types_to_enum.py
```diff
@@ -0,0 +1,61 @@
...
+
+"""
+
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = "69c7d2f49a4e"
+down_revision = "b31a27c8584f"
```
ErebusZ, line 15: do there have to be lower case?

elyousfi5: !!! this is the migration file, and its autogenerated !

3asm (approved):

PR #739, opened 2024-07-01
Persist the scan first then install agents when running a scan using the API

elyousfi5:
Run scan:
[Screencast from 2024-07-02 08-55-33.webm](https://github.com/Ostorlab/oxo/assets/144013278/0e31541d-790e-40b7-974c-c3a66bfdd026)

oxo scan list:
Screenshot from 2024-07-03 10-56-35
Screenshot from 2024-07-03 10-56-16


PiranhaSa (approved):
lgtm

src/ostorlab/runtimes/local/runtime.py
```diff
@@ -170,6 +170,23 @@ def _docker_checks(self):
 
         self._docker_client = docker.from_env()
 
+    def prepare_scan(
```
3asm, line 173: As discussed, the API should return once the scan is persisted the scan runtime should start in a dedicated thread or process.

3asm (changes requested):

src/ostorlab/runtimes/local/runtime.py
```diff
@@ -170,6 +170,23 @@ def _docker_checks(self):
...
+    ) -> None:
+        """Prepare scan entry in the database.
+
+        Args:
+            title: Scan title.
+            assets: The target asset to scan.
+        """
+
+        console.info("Creating scan entry")
+        if assets is None:
```
benyissa, line 184: I would add a unit test for this case.

benyissa (commented):
LGTM

src/ostorlab/runtimes/local/runtime.py
elyousfi5: which case man, this is a copy paste (refactoring)

tests/serve_app/oxo_test.py
```diff
@@ -2360,14 +2363,15 @@ def testRunScanMutation_whenNetworkAsset_shouldRunScan(
...
-    assert args["assets"][1].mask == "24"
+    assert prepare_scan_mock.called is True
+    title = prepare_scan_mock.call_args[1]["title"]
+    assets = prepare_scan_mock.call_args[1]["assets"]
+    assert title == "Test Scan Network Asset"
+    assert len(assets) == 2
+    assert assets[0].host == "8.8.8.8"
+    assert assets[0].mask == "32"
+    assert assets[1].host == "8.8.4.4"
+    assert assets[1].mask == "24"
```
oussamaessaji, line 2374: Assert on the scan id also

3asm:
The target is not showing in the video?

elyousfi5:
> The target is not showing in the video?

The video wasnt updated after the fix. Please check again.

src/ostorlab/runtimes/local/runtime.py
```diff
@@ -170,6 +170,24 @@ def _docker_checks(self):
...
+            assets_str = "N/A"
+        else:
+            assets_str = f'{", ".join([str(asset) for asset in assets])}'
```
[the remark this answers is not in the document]
3asm: Why do we need asset str if we have asset passed to it?

src/ostorlab/runtimes/local/runtime.py
```diff
@@ -190,13 +208,8 @@ def scan(
...
         try:
-            console.info("Creating scan entry")
-            if assets is None:
-                assets_str = "N/A"
-            else:
-                assets_str = f'{", ".join([str(asset) for asset in assets])}'
-                # TODO(mohsinenar): we need to add support for storing multiple assets and rename this to target.
-            self._scan_db = self._create_scan_db(asset=assets_str[:255], title=title)
+            if hasattr(self, "_scan_db") is False or self._scan_db is None:
+                self.prepare_scan(title=title, assets=assets)
```
3asm, line 212: Why do we need to check hasattr, this hint at a bug somewhere.

src/ostorlab/serve_app/oxo.py
```diff
@@ -793,6 +794,60 @@ def mutate(
 
         runtime_instance: runtime.LocalRuntime = runtime.LocalRuntime()
         runtime_instance.follow = []
+        created_scan = runtime_instance.prepare_scan(
+            assets=scan_assets, title=scan.title
+        )
+        RunScanMutation._persist_scan_info(created_scan, scan)
+
+        executor = futures.ThreadPoolExecutor(max_workers=1)
```
3asm, line 802: ThreadPoolExecutor for a single thread, why not use a single thread? or you can pool executor but assign that in the init and use it for all tasks.

src/ostorlab/serve_app/oxo.py
```diff
@@ -793,6 +794,60 @@ def mutate(
...
+            RunScanMutation._run_scan_background,
+            runtime_instance,
+            agent_group,
+            scan,
+            scan_assets,
+        )
+        return RunScanMutation(scan=created_scan)
+
+    @staticmethod
+    def _persist_scan_info(
```
3asm, line 813: Why static methods?

3asm (changes requested):

src/ostorlab/serve_app/oxo.py
elyousfi5: As discussed, these are specific to the mutation so they should be static

src/ostorlab/runtimes/local/runtime.py
elyousfi5: the init was mocked in some unit tests, i did address that.

3asm (approved):

PR #752, opened 2024-07-11
Don't show banner and startup logs

elyousfi5:
Before: 
image

After:
image


src/ostorlab/cli/serve/serve.py
```diff
@@ -19,6 +21,11 @@ def serve(ctx: click.core.Context, host: str, port: int, refresh_api_key: bool)
     """Run the Flask serve with the specified host and port."""
     try:
         from ostorlab.serve_app import app
+        import werkzeug
+
+        cli = sys.modules["flask.cli"]
+        cli.show_server_banner = lambda *x: None
+        werkzeug.serving.BaseWSGIServer.log_startup = lambda *x: None
```
mohsinenar, line 28: Seems like dirty trick! are you sure this is the right way to do it!!

mohsinenar (changes requested):

benyissa (approved):

src/ostorlab/cli/serve/serve.py
elyousfi5: do you have a better way to do it ?

3asm: I do, override the BaseWSGIServer, override the method and use the new instance. This is indeed a not so great hack.

3asm (changes requested):

src/ostorlab/cli/serve/serve.py
elyousfi5: We don't use it explicitly, its used by flask.

src/ostorlab/cli/serve/serve.py
```diff
@@ -19,6 +21,12 @@ def serve(ctx: click.core.Context, host: str, port: int, refresh_api_key: bool)
     """Run the Flask serve with the specified host and port."""
     try:
         from ostorlab.serve_app import app
+        import werkzeug
+
+        cli = sys.modules["flask.cli"]
+        cli.show_server_banner = lambda *x: None
+        # Disable werkzeug log_startup, this can't be done by overriding BaseWSGIServer class because it's used by Flask
```
3asm, line 28:
```suggestion
        # Disable werkzeug log_startup, this can't be done by overriding BaseWSGIServer class because it's used by Flask.
```

3asm (approved):

3asm (approved):

oussamaessaji (approved):

PR #753, opened 2024-07-12
Api schema file protos

adnaneserrar:


amine3 (changes requested):
The compilation was not done from the right folder it should be from ostorlab folder

src/ostorlab/agent/message/proto/v3/asset/file/api_schema/graphql/graphql.proto
```diff
@@ -0,0 +1,9 @@
+syntax = "proto2";
+
+package ostorlab.agent.message.proto.v3.asset.file.api_schema.graphql;
+
+message Message{
+  optional bytes content = 1;
+  optional string url = 2;
+  optional string type = 3;
```
elyousfi5, line 8: what is type ?

rabsondev (approved):

src/ostorlab/agent/message/proto/v3/asset/file/api_schema/graphql/graphql.proto
adnaneserrar: schema_type i'll change the name

benyissa (dismissed):
could u please Merge the latest changes from main into this branch.

src/ostorlab/agent/message/proto/v3/asset/file/api_schema/graphql/graphql.proto
elyousfi5: i meant why will it be used for ? from what i understood it can be graphql, wsdl etc right ? so no need since you have in proto name "graphql.proto"

3asm (changes requested):

3asm (approved):

PR #777, opened 2024-08-12
Trigger at_exit when an agent crashes

ErebusZ:


benyissa (commented):
I would call it within the finally block to ensure cleanup actions are always executed, regardless of success or failure of the process.

PR #779, opened 2024-08-12
Push Logs to GCP when starting a new scan. 

benyissa:


src/ostorlab/cli/console.py
```diff
@@ -45,6 +49,8 @@ def error(self, text: str) -> None:
         self._console.print(
             f":small_red_triangle: [bold]ERROR:[/] {text}", style="error"
```
adnaneserrar, line 50: isn't it better to use string literal

mohsinenar (approved):

PR #801, opened 2024-09-11
Fix emitting non-exact out selectors.

3asm:
`out-selectors` enable a hierarchical logic to receive and emit messages. If an agent list listens to `v3.report` then he will receive all messages like `v3.report`, `v3.report.vulnerability`, `v3.report.xxx`.

If an agent states `out-selector`, he should in the same fashion be able to emit all sub-messages.

The current implementation was doing an exact match:

```python
ostorlab.agent.agent.NonListedMessageSelectorError
```

Fix switch to `startswith` check:
```python
        if (
            any(
                out_selector.startswith(selector) for out_selector in self.out_selectors
            )
            is True
        ):
```.

Unit test checks for both should and should not emit case.

### DIFF
--- a/src/ostorlab/agent/agent.py
+++ b/src/ostorlab/agent/agent.py
@@ -372,10 +372,10 @@ class AgentMixin(
             None
         """
         if (
-            any(
-                out_selector.startswith(selector) for out_selector in self.out_selectors
+            all(
+                selector.startswith(out_selector) for out_selector in self.out_selectors
             )
-            is True
+            is False
         ):
             logger.error("selector not present in list of out selectors")
             # CAUTION: this check is enforced on the client-side only in certain runtimes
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by benyissa — DISMISSED** [review]

**Review by amine3 — APPROVED** [review]

File: src/ostorlab/agent/agent.py [hunk 1]
```diff
@@ -371,7 +371,12 @@ def emit_raw(
         Returns:
             None
         """
-        if selector not in self.out_selectors:
+        if (
+            any(
+                out_selector.startswith(selector) for out_selector in self.out_selectors
+            )
+            is True
+        ):
```
mohsinenar, line 379: This should be `is False`

mohsinenar (dismissed):

nmasdoufi-ol (approved):

PR #813, opened 2024-09-30
Add extra_headers to all v3.asset.file.api_schema protos

elyousfi5:
This PR is a part of adding API Credentials to agent graphql analyzer. It Adds extra_headers to v3.asset.file.api_schema protos.

for `pytest-httpx==0.30.0`, its pinned because it was recently released and the httpx_mock was not working.


src/ostorlab/agent/message/proto/v3/asset/file/api_schema/graphql/graphql.proto
```diff
@@ -9,8 +9,14 @@ message AndroidMetadata {
 message IOSMetadata {
   optional string bundle_id = 1;
 }
+
+message Header {
```
3asm, line 13: Should you add this to each `api_schema`. Not just GraphQL.

3asm (changes requested):

setup.cfg
```diff
@@ -101,7 +101,7 @@ testing =
     pytest-asyncio
     pytest-mock
     pytest-timeout
-    pytest-httpx
+    pytest-httpx==0.30.0
```
benyissa, line 104: Can you explain why we're pinning this version, in case we need to check later if an upgrade is possible?

3asm (approved):

PR #814, opened 2024-09-30
Add proto for kill scan message

oussamaessaji:
This PR implements the proto for kill message, this message will be used to kill all the agents within a universe.


src/ostorlab/agent/message/proto/v3/report/event/scan/kill/kill.proto
3asm, line 1: The selector should `v3.kill`.
Inside should be a string reason.

3asm (changes requested):

src/ostorlab/agent/message/proto/v3/report/event/scan/kill/kill.proto
benyissa: @3asm shouldn't we keep the same structure as scan.done ?

3asm (approved):

PR #834, opened 2024-10-25
Fix open ports parsing from yaml.

deadly-panda:
The `open_ports` field of the `AgentDefinition` class is of type `PortMapping`. 
When reading the raw yaml agent definition to create the  `AgentDefinition` instance, the current implementation doesn't convert the raw dict to the corresponding `PortMapping` objects.

This PR  handles this cases in the `from_yaml` classmethod of  the `AgentDefinition` class.

### DIFF
--- a/src/ostorlab/agent/definitions.py
+++ b/src/ostorlab/agent/definitions.py
@@ -54,7 +54,13 @@ class AgentDefinition:
             mounts=definition.get("mounts", []),
             restart_policy=definition.get("restart_policy", ""),
             mem_limit=definition.get("mem_limit"),
-            open_ports=definition.get("open_ports", []),
+            open_ports=[
+                defintions.PortMapping(
+                    source_port=p.get("src_port"),
+                    destination_port=p.get("dest_port"),
+                )
+                for p in definition.get("open_ports", [])
+            ],
             restrictions=definition.get("restrictions", []),
             version=definition.get("version"),
             description=definition.get("description"),
--- a/tests/agent/definitions_test.py
+++ b/tests/agent/definitions_test.py
@@ -3,6 +3,7 @@
 import io
 
 from ostorlab.agent import definitions
+from ostorlab.utils import defintions as utils_defintions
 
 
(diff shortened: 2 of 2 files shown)


rabsondev (approved):

PiranhaSa (approved):
:+1: :+1:

amine3 (approved):

Zakaria-Latif (approved):

tests/agent/definitions_test.py
```diff
@@ -42,3 +43,45 @@ def testAgentDefinitionFromYaml_whenYamlIsValid_returnsValidAgentDefinition():
...
+
+    yaml_data_file = io.StringIO(valid_yaml_data)
+
+    agent_definition = definitions.AgentDefinition.from_yaml(yaml_data_file)
+
+    assert agent_definition.name == "agent42"
+    assert agent_definition.caps == ["NET_ADMIN"]
+    assert agent_definition.open_ports == [
+        utils_defintions.PortMapping(
+            source_port=4242,
```
ybadaoui-ostorlab, line 84: i think you may want to test with different ports so you make sure everything is in the right place

src/ostorlab/agent/definitions.py
```diff
@@ -54,7 +54,13 @@ def from_yaml(cls, file: io.TextIOWrapper) -> "AgentDefinition":
             mounts=definition.get("mounts", []),
             restart_policy=definition.get("restart_policy", ""),
             mem_limit=definition.get("mem_limit"),
-            open_ports=definition.get("open_ports", []),
+            open_ports=[
+                defintions.PortMapping(
+                    source_port=p.get("src_port"),
+                    destination_port=p.get("dest_port"),
+                )
+                for p in definition.get("open_ports", [])
```
ohachimOs, line 62: shouldn't it be `p in definition.get("open_ports", {})`

ohachimOs: nvm :facepalm: :facepalm:

PR #842, opened 2024-12-02
Support Arguments with Colons in Their Values

benyissa:
### Summary

This PR fixes an issue with handling argument values containing colons (`:`) when constructing arguments. Previously, arguments with colons in their values were not processed correctly, leading to unexpected behavior. 

#### Before Fix:

Before Fix

#### After Fix:

After Fix

### DIFF
--- a/tests/cli/scan/run/run_test.py
+++ b/tests/cli/scan/run/run_test.py
@@ -391,6 +391,27 @@ def testOstorlabScanRunCLI_whenWrongArgsFormatProvided_showsErrorMessage() -> No
     )
 
 
+def testOstorlabScanRunCLI_whenAgentArgContainsThreeColons_doesNotCrash(
+    mocker: plugin.MockerFixture,
+) -> None:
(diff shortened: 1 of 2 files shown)


rabsondev (dismissed):

3asm:
@benyissa the crash is weird.

adnaneserrar (dismissed):

oussamaessaji (commented):
My concern is whether we really need to split the `:` in the first place.

tests/cli/scan/run/run_test.py
```diff
@@ -391,6 +391,28 @@ def testOstorlabScanRunCLI_whenWrongArgsFormatProvided_showsErrorMessage() -> No
     )
 
 
+def testOstorlabScanRunCLI_whenAgentArgContainsThreeColons_doesNotCrash(
```
ybadaoui-ostorlab, line 394:
```suggestion
def testOstorlabScanRunCLI_whenAgentArgContainsMultipleColons_doesNotCrash(
```

3asm (approved):

PR #849, opened 2025-01-03
Add option to change scan timeout in local scans

elyousfi5:
Add option to change scan timeout value in local scans.

### DIFF
--- a/src/ostorlab/cli/scan/run/run.py
+++ b/src/ostorlab/cli/scan/run/run.py
@@ -154,6 +154,8 @@ def run(
             console.error(f"{e}")
             raise click.ClickException("Invalid asset Group Definition.") from e
     runtime_instance: runtime.Runtime = ctx.obj["runtime"]
+    if timeout is not None:
+        runtime_instance.timeout = timeout
 
     # Prepare and set the list of agents to follow.
     agent_keys = [agent.key for agent in agent_group.agents]
@@ -185,7 +187,6 @@ def run(
                     title=ctx.obj["title"],
                     agent_group_definition=ctx.obj["agent_group_definition"],
                     assets=asset_group.targets if asset_group is not None else None,
-                    timeout=timeout,
                 )
                 if created_scan is not None:
                     runtime_instance.link_agent_group_scan(created_scan, agent_group)
--- a/src/ostorlab/cli/scan/run/run.py
+++ b/src/ostorlab/cli/scan/run/run.py
@@ -5,7 +5,7 @@ Example of usage:
 
 import io
 import logging
-from typing import List
+from typing import List, Optional
 
 import click
 import httpx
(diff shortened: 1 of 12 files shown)


tests/cli/scan/run/run_test.py
```diff
@@ -537,6 +537,94 @@ def testOstorlabScanRunCLI_whenFollow_shouldFollowSpecifiedAgents(
...
+        and arg.value == 0
+        and arg.type == "number"
+        for arg in mock_start_agent.call_args[1].get("agent").args
+    )
+    assert any(
+        arg.name == "scan_done_timeout_sec"
+        and arg.value == 3600
+        and arg.type == "number"
+        for arg in mock_start_agent.call_args[1].get("agent").args
+    )
```
rabsondev, line 590:
```suggestion
    assert any(
        arg.name == "postscane_done_timeout_sec"
        and arg.value == 0
        and arg.type == "number"
        for arg in mock_start_agent.call_args[1].get("agent").args
    ) is True
    assert any(
        arg.name == "scan_done_timeout_sec"
        and arg.value == 3600
        and arg.type == "number"
        for arg in mock_start_agent.call_args[1].get("agent").args
    ) is True
```

ohachimOs (dismissed):

oussamaessaji (dismissed):

src/ostorlab/cli/scan/run/run.py
```diff
@@ -101,6 +108,7 @@ def run(
     follow: List[str],
     no_follow: bool,
     no_asset: bool,
+    timeout: Optional[int] = None,
```
Zakaria-Latif, line 111:
```suggestion
    timeout: int | None = None,
```

src/ostorlab/runtimes/runtime.py
```diff
@@ -24,6 +24,7 @@ class Runtime(abc.ABC):
     """Runtime is in charge of preparing the environment to trigger a scan."""
 
     follow: list
+    timeout: Optional[int] = None
```
Zakaria-Latif, line 27:
```suggestion
    timeout: int | None = None
```

benyissa (dismissed):

src/ostorlab/cli/scan/run/run.py
elyousfi5: oxo has python3.9 in the workflows so this will fail.

nmasdoufi-ol (dismissed):

src/ostorlab/runtimes/local/runtime.py
```diff
@@ -519,7 +521,21 @@ def _list_agent_services(self):
...
+        )
+
+        if self.timeout is not None:
+            tracker_agent_settings.args.extend(
+                [
+                    utils_definitions.Arg(
+                        name="scan_done_timeout_sec", type="number", value=self.timeout
+                    ),
+                    utils_definitions.Arg(
+                        name="postscane_done_timeout_sec", type="number", value=0
```
amine3, line 535: Why do we set this one to 0?

amine3 (changes requested):

src/ostorlab/runtimes/local/runtime.py
amine3: But the purpose of the post scan done is that we give time to the agent to finish processing the last tasks (flush vulnerabilities) so setting to 0 will break that.

amine3 (approved):

PR #862, opened 2025-02-04
Add method to check if a key is already mapped in Redis

burogurama:
This PR adds a helper method `exists` in `AgentPersistMixin` to check if a key is already mapped in Redis. This is useful when the mapped value size is large and we just want to check without using it.

### DIFF
--- a/src/ostorlab/agent/mixins/agent_persist_mixin.py
+++ b/src/ostorlab/agent/mixins/agent_persist_mixin.py
@@ -113,7 +113,7 @@ class AgentPersistMixin:
         """
         return self._redis_client.get(key)
 
-    def exists(self, key: Union[bytes, str]) -> bool:
+    def exists(self, key: bytes | str) -> bool:
         """Helper function that checks if a given key exists in the Redis database.
 
         Args:
--- a/src/ostorlab/agent/mixins/agent_persist_mixin.py
+++ b/src/ostorlab/agent/mixins/agent_persist_mixin.py
@@ -113,6 +113,18 @@ class AgentPersistMixin:
         """
         return self._redis_client.get(key)
 
+    def exists(self, key: Union[bytes, str]) -> bool:
+        """Helper function that checks if a given key exists in the Redis database.
+
+        Args:
+            key: The key to check for existence.
+
+        Returns:
+            bool: True if the key exists, False otherwise.
+        """
+
+        return self._redis_client.exists(key) == 1
+
     def hash_add(
         self,
         hash_name: Union[bytes, str],
@@ -123,7 +135,7 @@ class AgentPersistMixin:
 
         Args:
             hash_name: Name of the hash.
-            mapping: Dict of key/value pairs that will be hadded to hash_name
+            mapping: Dict of key/value pairs that will be added to hash_name
 
         Returns:
             True if the key does not exist, False otherwise
--- a/src/ostorlab/agent/mixins/agent_persist_mixin.py
+++ b/src/ostorlab/agent/mixins/agent_persist_mixin.py
@@ -113,7 +113,7 @@ class AgentPersistMixin:
         """
         return self._redis_client.get(key)
 
-    def exists(self, key: bytes | str) -> bool:
+    def exists(self, key: Union[bytes, str]) -> bool:
         """Helper function that checks if a given key exists in the Redis database.
 
         Args:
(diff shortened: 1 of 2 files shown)


tests/agent/mixins/agent_persist_mixin_test.py
```diff
@@ -39,6 +39,25 @@ async def testAgentPersistMixin_whenSetIsAdded_setIsPersisted(
     assert mixin.value_type("myKey1") == "none"
 
 
+@pytest.mark.parametrize("clean_redis_data", ["redis://localhost:6379"], indirect=True)
+@pytest.mark.asyncio
+@pytest.mark.docker
+async def testAgentPersistMixinExists_whenKeyExists_returnTrue(
+    mocker, redis_service, clean_redis_data
+):
```
nmasdoufi-ol, line 47:
```suggestion
) -> None:
```

tests/agent/mixins/agent_persist_mixin_test.py
```diff
@@ -39,6 +39,25 @@ async def testAgentPersistMixin_whenSetIsAdded_setIsPersisted(
     assert mixin.value_type("myKey1") == "none"
 
 
+@pytest.mark.parametrize("clean_redis_data", ["redis://localhost:6379"], indirect=True)
+@pytest.mark.asyncio
+@pytest.mark.docker
+async def testAgentPersistMixinExists_whenKeyExists_returnTrue(
+    mocker, redis_service, clean_redis_data
```
mohsinenar, line 46: Why you use parametrize adn pass args just to delete them?

mohsinenar (changes requested):

src/ostorlab/agent/mixins/agent_persist_mixin.py
```diff
@@ -113,6 +113,18 @@ def get(self, key: Union[bytes, str]) -> Optional[bytes]:
         """
         return self._redis_client.get(key)
 
+    def exists(self, key: Union[bytes, str]) -> bool:
```
ohachimOs, line 116: Use '|'  :+1: 
```suggestion
    def exists(self, key: bytes |  str) -> bool:
```

tests/agent/mixins/agent_persist_mixin_test.py
ohachimOs: Hmmm this is what the existing unittests are doing, maybe it's because they are needed to setup something but not for the test? like the `clean_redis_data`
@nassermohamedit0

tests/agent/mixins/agent_persist_mixin_test.py
```diff
@@ -14,10 +14,10 @@
 @pytest.mark.asyncio
 @pytest.mark.docker
 async def testAgentPersistMixin_whenSetIsAdded_setIsPersisted(
-    mocker, redis_service, clean_redis_data
+    redis_service, clean_redis_data
```
oussamaessaji, line 17: add types please

src/ostorlab/agent/mixins/agent_persist_mixin.py
```diff
@@ -113,6 +113,18 @@ def get(self, key: Union[bytes, str]) -> Optional[bytes]:
...
+        """Helper function that checks if a given key exists in the Redis database.
+
+        Args:
+            key: The key to check for existence.
+
+        Returns:
+            bool: True if the key exists, False otherwise.
+        """
+
+        return self._redis_client.exists(key) == 1
```
3asm, line 126: According the the doc, a key can exist multiple times: https://redis.io/docs/latest/commands/exists/

I suggest keeping the API returning an int to align wit the behavior of the API.

3asm (changes requested):

src/ostorlab/agent/mixins/agent_persist_mixin.py
nassermohamedit0: Here we are checking for a single key and is not repeated in the arguments given to Redis client `exists` method so we are sure that it will return either one or zero.
Should I change the method signature to match its Redis client counterpart?  Or keep it as is and just return the the same value returned by Redis client `exists` method?

3asm: Thanks for the clarification. LGTM.

3asm (approved):

PR #863, opened 2025-02-13
Fix: Make asset classes fields dataclass fields

burogurama:
### Issue:

Every asset class is a `dataclass` (e.g., `File`, `AndroidStore`). This causes the `__eq__` method to be implicitly overridden.  

The `dataclass` implementation of `__eq__` compares only the fields explicitly defined as data fields. However, fields that are initialized inside `__init__` but not declared as data fields are ignored in equality comparisons.  

Example: `ostorlab.asset.IOSStore`

```python
@dataclasses.dataclass
@asset.selector("v3.asset.store.ios_store")
class IOSStore(asset.Asset):
    """Ios bundle target asset."""

    def __init__(self, bundle_id: str):
        self.bundle_id = bundle_id # Would not be used in ==
```

This causes two instances of `IOSStore` to be compared as equal (`==`) even if the `bundle_id` field is different. As a result, some tests may falsely succeed.

Example in `agent_cloud_persist_vulnz`:

```python
def testCloudPersistVulnzAgent_whenVulnerableTargetIsIOS_addVulnzCalled(
    mocker,
    persist_vulnz_agent_no_max_queue_obj,
    minimalist_report_ios_vulnerability_message,
):
    """Ensure that vulnerability location information is set correctly for an iOS store application."""
    call_request = mocker.patch(
        "agent.apis.call_add_vulnerabilities", return_value=None
    )
    vulnz = [
        apis.Vulnerability(
            ...
            vulnerability_location=vuln_mixin.VulnerabilityLocation(
                asset=ios_store.IOSStore(bundle_id="a.b.c"), # Test does not fail if you change the bundle_id value
                metadata=[
                    vuln_mixin.VulnerabilityLocationMetadata(
                        metadata_type=vuln_mixin.MetadataType.FILE_PATH,
                        value="dir/dummy.swift",
                    )
                ],
            ),
        )
    ]

    persist_vulnz_agent_no_max_queue_obj.process(
        minimalist_report_ios_vulnerability_message
    )

    call_request.assert_called_with("https://ostorlab.co/api", "my_token", 1, vulnz) # Would eventually copare two IOSStore instances
```

### Fix:

Make all fields of an asset class data fields.

### Update:

This PR fixes an existing unit test: `tests/runtimes/definitions_test::testAssetGroupDefinitionFromYaml_whenYamlIsValid_returnsValidAssetGroupDefinition`, which has a bug and started failing against the current PR (see diff).


src/ostorlab/assets/android_apk.py
```diff
@@ -11,6 +11,10 @@
 class AndroidApk(asset.Asset):
     """Android .APK  target asset."""
 
+    content: Optional[bytes] = None
```
ohachimOs, line 14: User '|'

adnaneserrar: +1

nassermohamedit0: This is no supported by Python 3.9
.

src/ostorlab/assets/agent.py
```diff
@@ -11,6 +11,12 @@
 class Agent(asset.Asset):
     """Agent asset."""
 
+    key: str
+    version: Optional[str] = None
+    git_location: Optional[str] = None
+    docker_location: Optional[str] = None
+    yaml_file_location: Optional[str] = None
```
nmasdoufi-ol, line 18: use | instead of Optional please

src/ostorlab/assets/ipv4.py
```diff
@@ -11,6 +11,10 @@
 class IPv4(asset.Asset):
     """IPv4 Address target asset."""
 
+    host: str
+    version: Optional[int] = 4
```
mohsinenar, line 15: Version can't be None, the typing in init was not correct!
```suggestion
    version: int = 4
```

nassermohamedit0: Done

elyousfi5 (commented):
This needs more testing
image

nassermohamedit0:
> This needs more testing image

I just added data fields in dataclasses. No new behavior.

nassermohamedit0:
> a comment

I just added data fields in dataclasses. No new behavior.

nassermohamedit0:
> This needs more testing image

I just added data fields in dataclasses. No new behavior.

benyissa (commented):
The @dataclass decorator automatically generates the __init__ method, __repr__, __eq__

@nassermohamedit0 could u please delete the `__init__` method in this case.

nassermohamedit0:
> The @DataClass decorator automatically generates the **init** method, **repr**, **eq**
> 
> @nassermohamedit0 could u please delete the `__init__` method in this case.

I wanted to be super sure not to break anything so I left the __init__ methods. But I think the explicit __init__ method are redundant.

src/ostorlab/assets/agent.py
```diff
@@ -11,6 +11,12 @@
 class Agent(asset.Asset):
     """Agent asset."""
 
+    key: str
```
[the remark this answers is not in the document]
nassermohamedit0: pass

deadly-panda (approved):

3asm (approved):

PR #868, opened 2025-02-23
Update agent_persist_mock to patch _exists method

burogurama:
This PR updates `persist_agent_mock` to patch the `_exists` method that was added to the `agent_persist_mixin`.

This PR also updates the patch function `_add` to only encode the `value` argument if it is not of type `bytes`.
If `value` is of type `bytes` then we have `str(value).encode() != value`.

### DIFF
--- a/src/ostorlab/testing/agent.py
+++ b/src/ostorlab/testing/agent.py
@@ -94,7 +94,7 @@ def agent_persist_mock(mocker):
         if key in storage:
             return storage[key]
 
-    def _add(key: Union[str, bytes], value: bytes):
+    def _add(key: Union[str, bytes], value: bytes) -> None:
         """Check values are present in the storage dict."""
         if isinstance(value, bytes) is False:
             value = str(value).encode()
--- a/src/ostorlab/testing/agent.py
+++ b/src/ostorlab/testing/agent.py
@@ -1,10 +1,9 @@
 """mock agent implements the required methods to test the agent's behavior without using external components."""
 
+from typing import List, Union
 import dataclasses
 
 import pytest
-
-from typing import List
 
 from ostorlab.agent.message import message as msg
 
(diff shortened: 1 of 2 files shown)


elyousfi5 (dismissed):

src/ostorlab/testing/agent.py
```diff
@@ -95,9 +94,15 @@ def _get(key):
         if key in storage:
             return storage[key]
 
-    def _add(key, value):
+    def _add(key: Union[str, bytes], value: bytes):
         """Check values are present in the storage dict."""
-        storage[key] = str(value).encode()
+        if isinstance(value, bytes) is False:
+            value = str(value).encode()
+        storage[key] = value
```
deadly-panda, line 101: Shouldn't this method return a boolean?

src/ostorlab/testing/agent.py
```diff
@@ -1,11 +1,10 @@
 """mock agent implements the required methods to test the agent's behavior without using external components."""
 
+from typing import List, Union
```
nmasdoufi-ol, line 3: please use list and | instead

src/ostorlab/testing/agent.py
nassermohamedit0: Looking at the source code, redis client `set` method does not return `bool`, it returns `Union[Awaitable[Any], Any]`. The `add` method in persist mixin re-returns this value so ig the return type (bool) is not correct.

src/ostorlab/testing/agent.py
```diff
@@ -95,9 +94,15 @@ def _get(key):
         if key in storage:
             return storage[key]
 
-    def _add(key, value):
+    def _add(key: Union[str, bytes], value: bytes):
```
oussamaessaji, line 97: Add return type

3asm (approved):

PR #870, opened 2025-02-25
Sort dna dict

oussamaessaji:
In this PR, I've added the logic to sort the DNA dictionary to unify the sorting logic in one place and remove duplicate code, so the agent does not have to handle sorting on their side.

### DIFF
--- a/src/ostorlab/agent/mixins/agent_report_vulnerability_mixin.py
+++ b/src/ostorlab/agent/mixins/agent_report_vulnerability_mixin.py
@@ -6,6 +6,7 @@ and emit a vulnerability message.
 
 import dataclasses
 import enum
+import json
 from typing import Optional, List, Any, Union, Dict
 
 from ostorlab.agent.kb import kb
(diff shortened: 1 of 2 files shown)


src/ostorlab/agent/mixins/agent_report_vulnerability_mixin.py
```diff
@@ -145,3 +151,22 @@ def report_vulnerability(
...
+
+
+def _sort_dict(dictionary: dict[str, Any] | list[Any]) -> dict[str, Any] | list[Any]:
+    """Recursively sort dictionary keys and lists within.
+    Args:
+        dictionary: The dictionary to sort.
+    Returns:
+        A sorted dictionary or list.
+    """
+    if isinstance(dictionary, dict):
```
nmasdoufi-ol, line 163:
```suggestion
    if isinstance(dictionary, dict) is True:
```

Zakaria-Latif: mypy does not accept it

src/ostorlab/agent/mixins/agent_report_vulnerability_mixin.py
```diff
@@ -145,3 +151,22 @@ def report_vulnerability(
...
+        return {k: _sort_dict(v) for k, v in sorted(dictionary.items())}
+    if isinstance(dictionary, list):
```
Oussama-Bakri, line 165: if isinstance(dictionary, list) is True:

PR #876, opened 2025-03-05
Add record_type to the metadata type

MouadAO:
This pull request adds the record type, which is necessary to construct the vulnerability location for dns agent.


src/ostorlab/agent/message/proto/v3/report/vulnerability/vulnerability.proto
```diff
@@ -41,6 +41,7 @@ enum MetadataTypeEnum {
   CALL_TRACE = 8;
   VERSION = 9;
   INSERTION_POINT = 10;
+  RECORD_TYPE = 11;
```
ErebusZ, line 44: I think it would be clearer as `DNS_RECORD`

nmasdoufi-ol (dismissed):

src/ostorlab/agent/message/proto/v3/report/vulnerability/vulnerability.proto
benyissa, line 44:
```suggestion
  DNS_RECORD_TYPE = 11;
```

benyissa (approved):

src/ostorlab/agent/message/proto/v3/report/vulnerability/vulnerability.proto
MouadAO: we agreed on DNS_RECORD_TYPE

3asm (approved):

PR #889, opened 2025-03-26
feature: Add exploitation and post exploitation details

Zakaria-Latif:
Add exploitation and post exploitation details in vulnerability model


src/ostorlab/runtimes/local/models/models.py
```diff
@@ -269,6 +271,8 @@ def create(
         recommendation: str,
         technical_detail: str,
         risk_rating: str,
+        exploitation_detail: Optional[str] = None,
+        post_exploitation_detail: Optional[str] = None,
```
nmasdoufi-ol, line 275:
```suggestion
        exploitation_detail: str | None = None,
        post_exploitation_detail: str | None = None,
```

Zakaria-Latif: I tried it previously, it does not work in oxo

tests/runtimes/local/models/models_test.py
```diff
@@ -49,6 +49,8 @@ def testModelsVulnerability_whenDatabaseDoesNotExist_DatabaseAndScanCreated(
         description="Javascript Vuln",
         recommendation="Sanitize data",
         technical_detail="a=$input",
+        exploitation_detail="exploitation",
+        post_exploitation_detail="post exploitation",
```
rabsondev, line 53: Add assertions on these fields.

src/ostorlab/runtimes/local/models/models.py
deadly-panda: Oxo still supports python3.9. The `|` syntax was added in 3.10

elyousfi5 (approved):

PR #894, opened 2025-03-27
feature: Add universe id to all services

oussamaessaji:
In this PR, I've included the runtime_name (universe ID) for all services, as it was previously added only to the images that lacked the service_name attribute in the agent definition.

### DIFF
--- a/src/ostorlab/runtimes/lite_local/agent_runtime.py
+++ b/src/ostorlab/runtimes/lite_local/agent_runtime.py
@@ -353,9 +353,8 @@ class AgentRuntime:
         service_name = (
             agent_definition.service_name
             or self.agent.container_image.split(":")[0].replace(".", "")
-            + "_"
-            + self.runtime_name
         )
+        service_name = service_name + "_" + self.runtime_name
 
         # We apply the random str only if it will not break the max docker service name characters (63)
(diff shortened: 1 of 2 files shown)


src/ostorlab/runtimes/lite_local/agent_runtime.py
```diff
@@ -353,9 +353,8 @@ def create_agent_service(
         service_name = (
             agent_definition.service_name
             or self.agent.container_image.split(":")[0].replace(".", "")
-            + "_"
-            + self.runtime_name
         )
+        service_name = service_name + "_" + self.runtime_name
```
dahoth, line 357:
```suggestion
        service_name = f"{service_name}_{self.runtime_name}"
```

src/ostorlab/runtimes/local/agent_runtime.py
```diff
@@ -342,9 +342,8 @@ def create_agent_service(
         service_name = (
             agent_definition.service_name
             or self.agent.container_image.split(":")[0].replace(".", "")
-            + "_"
-            + self.runtime_name
         )
+        service_name = service_name + "_" + self.runtime_name
```
dahoth, line 346:
```suggestion
        service_name = f"{service_name}_{self.runtime_name}"
```

PR #916, opened 2025-09-10
feature: Add support tracker init_sleep argument to be controlled by the cli.

amine3:
Add support tracker init_sleep argument to be controlled by the CLI.

This is needed when the universe needs to wait for a long warmup before the messages get emitted.


burogurama (approved):
LGTM

3asm (approved):

PR #928, opened 2025-10-01
fix: when service name is defined in the agent definition, do not add random number

elyousfi5:
This PR fixes the service naming logic for agents to respect custom service names defined in agent definitions. When a service name is explicitly provided in the agent definition, now we uses it directly without appending random numbers, preventing naming conflicts and ensuring predictable service names.

Key changes:
- Modified service name generation logic to check for predefined service names first
- Added truncation logic for service names that exceed Docker's 63-character limit
- Added comprehensive test coverage for the new service naming behavior

### DIFF
--- a/src/ostorlab/runtimes/lite_local/agent_runtime.py
+++ b/src/ostorlab/runtimes/lite_local/agent_runtime.py
@@ -350,16 +350,21 @@ class AgentRuntime:
         )
         caps = self.agent.caps or agent_definition.caps
 
-        service_name = (
-            agent_definition.service_name
-            or self.agent.container_image.split(":")[0].replace(".", "")
-            + "_"
-            + self.runtime_name
-        )
+        if agent_definition.service_name is not None:
+            # If the service name defined in the agent definition is longer than the max allowed service name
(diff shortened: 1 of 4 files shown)


src/ostorlab/runtimes/lite_local/agent_runtime.py
```diff
@@ -350,16 +350,21 @@ def create_agent_service(
...
-        service_name = (
-            agent_definition.service_name
-            or self.agent.container_image.split(":")[0].replace(".", "")
-            + "_"
-            + self.runtime_name
-        )
+        if agent_definition.service_name is not None:
+            # If the service name defined in the agent definition is longer than the max allowed service name
+            # we truncate it to avoid docker api error.
+            service_name = agent_definition.service_name[:MAX_SERVICE_NAME_LEN]
```
3asm, line 356: Trucation of the provided service name is not expected, I would rather have an error at scan creation rather than this behavior.

3asm (changes requested):

3asm (approved):

amat-osto (approved):

PR #930, opened 2025-10-02
fix: adding test for risk field in vuln in protobuf

SalwaAdlouni:
This pull request adds a new unit test for the serialization and deserialization of vulnerability messages, specifically verifying that the `risk` field is correctly handled.

Testing improvements:

* Added a unit test `testVulnerabilityToProto_whenVulnerabilityWithRiskField_generatesProto` in `tests/assets/vulnerability_test.py` to ensure that vulnerability messages with a `risk` field are properly serialized and deserialized, and all expected fields are correctly mapped.

### DIFF
--- a/tests/assets/vulnerability_test.py
+++ b/tests/assets/vulnerability_test.py
@@ -6,33 +6,35 @@ from ostorlab.agent.message import serializer
 def testVulnerabilityToProto_whenVulnerabilityWithRiskField_generatesProto():
     """Test serializing and deserializing vulnerability message with risk field."""
     vulnerability_data = {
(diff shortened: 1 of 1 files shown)


burogurama (approved):

tests/assets/vulnerability_test.py
```diff
@@ -0,0 +1,40 @@
+"""Unit tests for Vulnerability serialization."""
```
3asm, line 1: This is not the right location. It should be under the same path as the proto or in the serializer tests.

3asm (changes requested):

PR #953, opened 2025-11-11
fix: service log stream threads stdout lock error after interpreter shutdown

burogurama:
This PR fixes service log stream implementation to ensure graceful shutdown of service log threads.

### Context:
When a scan is run, it follows the logs of services (unless --no-follow is set). After all services are started, a (daemon) thread is created for each service to stream its logs to the console. Another thread, let''s call it `M`, is started that checks statuses of the services. When all services are removed, it calls `sys.exit(0)` (Calling `sys.exit(0)` is actually useless because M is not the main thread, but since M outlives the main thread and usually it is the last thread standing, when it exist, the interpreter starts shutting down.)

### Problem:
It is possible that thread `M` exits while some logging threads are still running. Since logging threads are daemon Python does not wit for them and they are stoped abruptly. This can happen if the last service finishes but there are still logs that have not been processed by some logging threads. In this situation, the daemon thread that tries to write to stdout will trigger some error or undefined behavior because it is kinda a race condition.
In pentest subscans, this issue manifests frequently, see error blow:

```
Fatal Python error: _enter_buffered_busy: could not acquire lock for <_io.BufferedWriter name='<stdout>'> at interpreter shutdown, possibly due to daemon threads
Python runtime state: finalizing (tstate=0x000073b7a24aa6d0)

Current thread 0x000073b7a1b84b80 (most recent call first):
  <no Python frame>
```

### Fix:
The `logs` stream returned by the Docker SDK does not provide any notification when a service is removed or when no further logs will follow. It simply blocks indefinitely. Therefore, we still need to poll the service statuses. When a service is removed, we notify the corresponding logging thread via an event to make it stop. Because the thread may be blocking (for example, if all service logs have already been processed or if the underlying connection is slow), we call `join(timeout)` to give it a chance to exit gracefully. If, after the timeout, the thread has not checked the event, it is highly likely that it is blocked on an empty stream, which should not cause any issues during interpreter shutdown. Python has no mechanism to interrupt a thread and the docker SDK has no non-blocking or timeout options for this usecase.

### Changes:
- Refactor `logs_streamer` module to make implementation cleaner.
- Add the stop event logic to logging threads.
- Add `wait` method to `LogStream` to wait for all the service logs to be processed.
- Remove explicit `sys.exit(0)` because it is useless.
- Add unit tests for `log_streamer` module.


Oussama-Bakri (dismissed):
LGTM

MouadAO (dismissed):
LGTM

3asm (changes requested):

src/ostorlab/runtimes/local/log_streamer.py
```diff
@@ -68,3 +105,10 @@ def _select_color(self, service):
             color = COLOR_POOL.pop()
             COLOR_POOL.insert(0, color)
         return color
+
+    def _service_exist(self, service_id) -> bool:
```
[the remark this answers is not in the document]
nassermohamedit: Addressed.

src/ostorlab/runtimes/local/runtime.py
```diff
@@ -115,11 +114,12 @@ def __init__(
         self._mq_service: Optional[mq.LocalRabbitMQ] = None
         self._redis_service: Optional[redis.LocalRedis] = None
         self._jaeger_service: Optional[jaeger.LocalJaeger] = None
-        self._log_streamer = log_streamer.LogStream()
         self._scan_db: Optional[models.Scan] = None
         self._mq_exposed_ports: Optional[Dict[int, int]] = mq_exposed_ports
         self._gcp_logging_credential = gcp_logging_credential
         self._run_default_agents: bool = run_default_agents
+        self._log_streamer: Optional[log_streamer.LogStream] = None
```
SalwaAdlouni: avoid optional

SalwaAdlouni (approved):

3asm (approved):

src/ostorlab/runtimes/local/log_streamer.py
```diff
@@ -26,38 +28,73 @@
...
+            console.info(f"[{self._color} bold]{name}:[/] {line[:-1].decode()}")
+
+    def stop(self) -> None:
+        """Stop the log stream."""
+        if self._thread is None:
+            return
+        self._stop_event.set()
+        # thread may check the stop event late when the thread calling LogStream.wait has already exited
+        # so we join the thread with a timeout.
+        self._thread.join(timeout=1)
```
ohachimOs: One sec is enough?

PR #954, opened 2025-11-12
fix: oxo stop method

burogurama:
Error:
```
 File "/usr/local/lib/python3.12/site-packages/click/decorators.py", line 34, in new_func
   return f(get_current_context(), *args, **kwargs)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 File "/usr/local/lib/python3.12/site-packages/ostorlab/cli/scan/stop/stop.py", line 52, in stop
   runtime_instance.stop(scan_id=scan_id)
 File "/usr/local/lib/python3.12/site-packages/ostorlab/runtimes/local/runtime.py", line 304, in stop
   if service_labels.get("ostorlab.universe") is not None and int(
                                                              ^^^^
ValueError: invalid literal for int() with base 10: 'a_286565_162776'
```

The above failure occurs in prod. Some service has `a_286565_162776` as its universe, and the stop method fails because it tries to parse the universe as an int. The source of this unexpected value is yet to be determined. The issue also occurs locally.

This PR fixes the stop method so that it does not fail if a some random service has a non integer universe.

Changes:
- Fix typing, the stop method typing is incorrect, it is called with scan_id as `int` not a `str`.
- Remove parsing services universes, simply compare the universe as is with `str(scan_id)`


src/ostorlab/runtimes/local/runtime.py
```diff
@@ -278,18 +278,13 @@ def _check_services_running(self) -> None:
                 stop_event.set()
         sys.exit(0)
 
-    def stop(self, scan_id: str) -> None:
-        """Remove a service (scan) belonging to universe with scan_id(Universe Id).
+    def stop(self, scan_id: int) -> None:
```
3asm: How come mypy didn't flag this. Can you please ensure it covered here.

3asm (changes requested):

ohachimOs (approved):
No Unittests, something that would have broke previously but works now
Otherwise, LGTM

PR #974, opened 2026-01-19
fix: prioritize api key over token

Oussama-Bakri:
when api key is provided and user is authenticated with oxo at the same time, the authorization token will be used instead of provided api key, this PR fix this issue by prioritizing api key over authorization token


3asm (approved):

yasser-abassi (approved):

SalwaAdlouni (approved):
LGTM

PR #985, opened 2026-03-02
feature: Add Ability To Specify Service Name In Agent Group

ohachimOs:
## Issue

In the AI Pentest agent group, agents like Crawler need to run as both an MCP server (used by Auto Exploit) and a standard OXO bus agent (emitting request/response to Threat Intel). The natural solution is two entries in the agent group YAML with different args, but both resolve to the same Docker service name baked into the image label, causing:

```
docker.errors.APIError: 409 Client Error: Conflict ("service crawler already exists")
```

image

## Fix

Added a `service_name` field to agent group YAML entries. When set, it overrides the service name from the image label, allowing two entries with the same `key` to run as separate Docker services with independent configs:

```yaml
- key: agent/ostorlab/crawler
  args:
    - name: should_start_mcp_server
      type: boolean
      value: true

- key: agent/ostorlab/crawler
  service_name: crawler_bus
```

image

Fallback always unique: in the auto-generated service name path (no explicit service_name), the random suffix is now always appended — previously it was skipped when the base name was too long, causing collisions between multiple instances of the same agent; the base is now truncated to make room for the suffix instead

**Files changed:** `runtimes/definitions.py`, `runtimes/local/agent_runtime.py`, `runtimes/lite_local/agent_runtime.py`

## Note

Still fixing Cloud runtime.

### DIFF
--- a/src/ostorlab/runtimes/lite_local/agent_runtime.py
+++ b/src/ostorlab/runtimes/lite_local/agent_runtime.py
@@ -354,12 +354,13 @@ class AgentRuntime:
         )
         caps = self.agent.caps or agent_definition.caps
 
-        if agent_definition.service_name is not None:
-            if len(agent_definition.service_name) > MAX_SERVICE_NAME_LEN:
+        resolved_service_name = self.agent.service_name or agent_definition.service_name
+        if resolved_service_name is not None:
+            if len(resolved_service_name) > MAX_SERVICE_NAME_LEN:
                 raise ServiceNameTooLong(
-                    f'service name "{agent_definition.service_name}" exceeds max length of {MAX_SERVICE_NAME_LEN}'
+                    f'service name "{resolved_service_name}" exceeds max length of {MAX_SERVICE_NAME_LEN}'
                 )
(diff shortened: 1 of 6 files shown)


yami-ost (dismissed):
LGTM

MouadAO (dismissed):
LGTM

src/ostorlab/runtimes/lite_local/agent_runtime.py
```diff
@@ -354,12 +354,13 @@ def create_agent_service(
         )
         caps = self.agent.caps or agent_definition.caps
 
-        if agent_definition.service_name is not None:
-            if len(agent_definition.service_name) > MAX_SERVICE_NAME_LEN:
+        resolved_service_name = self.agent.service_name or agent_definition.service_name
```
3asm: The order of priority here is not explicit or clear. I suggest a more explicit code that shows which one take precedence.

3asm (changes requested):

3asm (approved):

PR #993, opened 2026-03-24
feature: Add retry to download docker image.

3asm:
Image download can fail mid-download for network issues and the whole flow will stop. 

The PR adds simple retry to ensure we try downloading in case of an error and also improve the error message.


MouadAO (approved):
LGTM

PR #998, opened 2026-04-09
fix: Add HarmonyOS proto metadata support and missing package init files

1l0c:
This PR introduces support for HarmonyOS protobuf metadata and fixes a runtime deserialization issue caused by missing package discovery.

#### **Changes**

* Add HarmonyOS metadata support to protobuf definitions
* Add missing `__init__.py` files for HarmonyOS proto packages to ensure proper package resolution

#### **Fix**

Resolves the following runtime error during message processing:

```
NoMatchingPackageNameError
```

This error occurred because the serializer could not resolve the protobuf package name due to missing package initialization files, preventing proper deserialization of messages.

#### **Root Cause**

* HarmonyOS proto modules were not recognized as Python packages
* `_find_package_name` failed to locate a matching package during deserialization


src/ostorlab/agent/message/proto/v3/asset/file/api_schema/graphql/graphql.proto
```diff
@@ -22,8 +26,9 @@ message Message{
   oneof metadata_oneof {
     AndroidMetadata android_metadata = 4;
     IOSMetadata ios_metadata = 5;
+    HarmonyOSMetadata harmonyos_metadata = 6;
```
amine3, line 29: No you should not do that add it always to the end to not break exiting messages

amine3 (changes requested):

src/ostorlab/agent/message/proto/v3/asset/file/api_schema/wsdl/wsdl.proto
```diff
@@ -22,8 +26,10 @@ message Message{
   oneof metadata_oneof {
     AndroidMetadata android_metadata = 4;
     IOSMetadata ios_metadata = 5;
+    HarmonyOSMetadata harmonyos_metadata = 6;
```
amine3, line 29: It applies to all the protos add always the change to the latest value so we do not break backward compatibility

amine3 (changes requested):

PR #1004, opened 2026-04-13
fix: bump python version to 3.14

mouadostor:
## Changes

- upgrade python version to 3.14


.github/workflows/lint_format_checker.yaml
```diff
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [ "3.9", "3.10" ]
+        python-version: [ "3.9", "3.10", "3.14" ]
```
3asm, line 14: 3.14 and 3.13. Drop the rest.

3asm (changes requested):

src/ostorlab/agent/mixins/agent_open_telemetry_mixin.py
```diff
@@ -47,15 +47,15 @@ def get_trace_exporter(self) -> sdk_export.SpanExporter:
...
+            eg: otlp://otel-collector-host:4317
             for gcp the format is:
             name_of_the_tracing_tool://project_id/service_account_json_base64
             eg: gcp://project_1/service_account_json_base64
         """
         parsed_url = parse.urlparse(self._tracing_collector_url)
         scheme = parsed_url.scheme
-        if scheme == "jaeger":
-            return self._get_jaeger_exporter(parsed_url)
+        if scheme in ("otlp", "jaeger"):
```
3asm, line 57: This code should be checked with the following documentation link: https://oxo.ostorlab.co/tutorials/debugging_agents#tracing

MouadAO (approved):
LGTM

setup.cfg
```diff
@@ -50,7 +50,7 @@ install_requires =
     docker
     importlib-metadata
     jsonschema>=4.4.0
-    protobuf==5.29.5
+    protobuf==6.33.6
```
1l0c, line 53: Did you test with other agents like the download agent? It may break. I previously had an issue with protobuf in the HarmonyOS download agent because I used a newer protoc and had to regenerate it with an older version. With this change, it might break again.

3asm (approved):

PR #1016, opened 2026-04-29
feature: Add missing startAgentScan attributes

ohachimOs:
## Fix `from_bus_message` proto field misalignment and missing `open_ports`/`service_name`

The oxo scanner proto (`_location/startAgentScan.proto`) was out of sync with the scanning engine proto. Field 8 was `accepted_agents` in oxo but `in_selectors` in SE, so in_selectors values were silently landing in the wrong field. `open_ports` and `service_name` were missing entirely, so port mappings were never published when using the `ostorlab scanner` CLI path (on-prem scanners). This caused agent services that expose ports (e.g. `agent_proxy`) to be unreachable from outside the container.

Updated the proto to match SE's field numbering, regenerated `_pb2.py`, and updated `from_bus_message` to handle all fields. Only affects the `ostorlab scanner` CLI path.


MouadAO (approved):
LGTM 👏

amat-osto (approved):

burogurama (approved):

Zakaria-Latif (approved):

PR #1021, opened 2026-05-05
feature: Feature/adding disk usage scanning status

zoulatizakaria:
Add disk usage tracking to scanner state

Added disk_usage field to ScannerState dataclass as a percentage (float, defaults to 0.0)
Added total_disk field to ScannerState dataclass as total disk size in GB (int, defaults to 0)
Populated both fields in ScannerStateReporter._capture_state using psutil.disk_usage("/")
Included diskUsage and totalDisk in the GraphQL mutation variables sent to the backend
Updated the scanner state reporter test to mock psutil.disk_usage and assert the correct values for both fields

### DIFF
--- a/src/ostorlab/utils/scanner_state_reporter.py
+++ b/src/ostorlab/utils/scanner_state_reporter.py
@@ -34,6 +34,7 @@ class ScannerStateReporter:
                 hostname=self._hostname,
                 ip=self._ip,
                 disk_usage=psutil.disk_usage("/").percent,
+                total_disk=psutil.disk_usage("/").total >> 30,  # total disk in GB
             )
         except ImportError:
             state = definitions.ScannerState(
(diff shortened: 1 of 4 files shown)


src/ostorlab/utils/scanner_state_reporter.py
```diff
@@ -33,6 +33,7 @@ def _capture_state(self) -> definitions.ScannerState:
                 total_memory=psutil.virtual_memory().total >> 30,  # total memory in GB
                 hostname=self._hostname,
                 ip=self._ip,
+                disk_usage=psutil.disk_usage("/").percent,
```
amine3, line 36: We need to add also the space because 80% of 1Tb is not the same as 40% of 10Gb disk

amine3 (changes requested):

3asm (approved):

PR #1030, opened 2026-05-12
fix: add ticket_key to ticket proto, asset, and scan run CLI

burogurama:
## Summary
- Adds the optional `ticket_key` field to the `v3.report.ticket` proto message (and regenerates `ticket_pb2.py`).
- Surfaces it on the `Ticket` asset dataclass and as a new `--ticket-key` option on `ostorlab scan run ticket`, so the value is serialized to the proto via the asset's existing `to_proto`.
- Extends the asset and CLI unit tests; proto-level test already covered the new field.


amat-osto (approved):
LGTM

ohachimOs (dismissed):

ohachimOs (approved):

PR #1052, opened 2026-06-02
fix: add event loop guard in agentMixin

HamzaElyoussfi:
Fix `AgentMQMixin` initialization when no current asyncio event loop is set.

`AgentMQMixin.__init__()` previously called `asyncio.get_event_loop()` directly when no loop was provided. In full test runs, earlier async tests can leave the main thread without a current loop, causing sync MQ tests to fail with `RuntimeError: There is no current event loop in thread 'MainThread'`.

This change aligns `AgentMQMixin` with the existing guard already used in `Agent`:
- use the provided loop when available
- otherwise try `asyncio.get_event_loop()`
- if no loop exists, create one with `asyncio.new_event_loop()` and register it with `asyncio.set_event_loop()`

This keeps the fix narrowly scoped to MQ mixin initialization and makes direct `AgentMQMixin` usage safe in synchronous contexts.

### DIFF
--- a/src/ostorlab/agent/mixins/agent_mq_mixin.py
+++ b/src/ostorlab/agent/mixins/agent_mq_mixin.py
@@ -43,7 +43,14 @@ class AgentMQMixin:
         self._queue_name = f"{self._name}_queue"
         self._url = url
         self._topic = topic
-        self._loop = loop or asyncio.get_event_loop()
+        if loop is not None:
+            self._loop = loop
+        else:
+            try:
+                self._loop = asyncio.get_event_loop()
+            except RuntimeError:
+                self._loop = asyncio.new_event_loop()
+                asyncio.set_event_loop(self._loop)
         self._max_priority = max_priority
         self._executor = concurrent.futures.ThreadPoolExecutor(max_workers=3)
         self._queue: Optional[aio_pika.Queue] = None
--- a/tests/agent/mixins/agent_mq_mixin_test.py
+++ b/tests/agent/mixins/agent_mq_mixin_test.py
@@ -155,6 +155,25 @@ def testMqSendMessage_onConnectionResetError_shouldRetriesAndReraise(
     assert mock_send_message.call_count == 6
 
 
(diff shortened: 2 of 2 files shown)


tests/agent/mixins/agent_mq_mixin_test.py
```diff
@@ -155,6 +155,25 @@ def testMqSendMessage_onConnectionResetError_shouldRetriesAndReraise(
     assert mock_send_message.call_count == 6
 
 
+def testAgentMqMixin_whenNoCurrentLoop_shouldCreateEventLoop(
```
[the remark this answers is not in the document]
HamzaElyoussfi: The required test is added

ohachimOs (approved):

src/ostorlab/agent/mixins/agent_mq_mixin.py
```diff
@@ -43,7 +43,14 @@ def __init__(
...
         self._topic = topic
-        self._loop = loop or asyncio.get_event_loop()
+        if loop is not None:
+            self._loop = loop
+        else:
+            try:
+                self._loop = asyncio.get_event_loop()
+            except RuntimeError:
+                self._loop = asyncio.new_event_loop()
+                asyncio.set_event_loop(self._loop)
```
yami-ost, line 53: Two minor, non-blocking notes on this block:

- **Shared helper:** this guard is now duplicated verbatim from `Agent.__init__` (`agent.py:107-112`). Worth extracting into a small util (e.g. `get_or_create_event_loop()`) so the two copies don't drift.
- **Deprecation:** `asyncio.get_event_loop()` (line 50) is deprecated since Python 3.12 when no loop is running. Out of scope here since this mirrors the existing `Agent` pattern, but both call sites will eventually want migrating.

HamzaElyoussfi: is still usable 

Yes its duplicated in Agent but Agentmixin is inherited by all other agent they wont need to recvheck

PR #1075, opened 2026-06-26
feature: Add support for inject asset args

MouadAO:
feature: Add support for inject asset args

### Key changes:
Previously Inject asset was a default agent oxo starts by default , and wasn't accepting args
currently we changed the logic so we can pass args in the agentGroup like this:

the oxo command:
```
/home/aitougrram/projects/agents/oxo/.venv/bin/oxo scan run -g mouad.yaml repository --repository-url https://bitbucket.org/mouad-osto/first_repo.git --commit-hash b3bcad4 --provider gitbucket
```

the agent group:

```
agents:
- args:
  - {name: nvd_api_key, type: string, value: XXXX-XXXX}
  caps: []
  cyclic_processing_limit: 0
  depth_processing_limit: 0
  in_selectors: [v3.fingerprint.file]
  key: agent/ostorlab/osv
  replicas: 1
- args:
  - {name: api_reporting_engine_base_url, type: string, value: 'http://4399.17.0.1:8002/apis/graphql'}
  - {name: reporting_engine_api_key, type: string, value: "XXXXX.XXXX"}
  caps: []
  cyclic_processing_limit: 0
  depth_processing_limit: 0
  in_selectors: []
  key: agent/ostorlab/inject_asset  # <--- Now we can do this and it won't result duplicate inject asset agent s 
  replicas: 1
description: Agent group definition for scan XXXX
kind: AgentGroup
name: test_agent_group

```

Screenshot from 2026-06-26 15-37-22

### DIFF
--- a/src/ostorlab/runtimes/lite_local/runtime.py
+++ b/src/ostorlab/runtimes/lite_local/runtime.py
@@ -195,7 +195,18 @@ class LiteLocalRuntime(runtime.Runtime):
                 raise AgentNotHealthy()
             if assets is not None:
                 console.info("Injecting assets")
-                self._inject_assets(assets=assets)
+                inject_asset_agent_settings = next(
+                    (
+                        agent
+                        for agent in agent_group_definition.agents
+                        if agent.key == ASSET_INJECTION_AGENT_DEFAULT
+                    ),
+                    None,
+                )
+                self._inject_assets(
+                    assets=assets,
+                    inject_asset_agent_settings=inject_asset_agent_settings,
+                )
         except AgentNotHealthy:
             console.error("Agent not starting")
             self.stop(self.scan_id)
@@ -270,6 +281,7 @@ class LiteLocalRuntime(runtime.Runtime):
             future_to_agent = {
                 executor.submit(self._start_agent, agent, extra_configs=[]): agent
                 for agent in agent_group_definition.agents
+                if agent.key != ASSET_INJECTION_AGENT_DEFAULT
             }
             for future in futures.as_completed(future_to_agent):
                 future.result()
@@ -344,7 +356,11 @@ class LiteLocalRuntime(runtime.Runtime):
             if service.name.startswith("agent_"):
                 yield service
 
-    def _inject_assets(self, assets: List[base_asset.Asset]):
+    def _inject_assets(
+        self,
+        assets: list[base_asset.Asset],
+        inject_asset_agent_settings: definitions.AgentSettings | None = None,
+    ):
         """Injects the scan target assets."""
 
         contents = {}
@@ -354,9 +370,11 @@ class LiteLocalRuntime(runtime.Runtime):
 
         volumes.create_volume(f"asset_{self.name}", contents)
 
-        inject_asset_agent_settings = definitions.AgentSettings(
-            key=ASSET_INJECTION_AGENT_DEFAULT, restart_policy="none"
-        )
+        if inject_asset_agent_settings is None:
+            inject_asset_agent_settings = definitions.AgentSettings(
+                key=ASSET_INJECTION_AGENT_DEFAULT, restart_policy="none"
+            )
+
         self._start_agent(
             agent=inject_asset_agent_settings,
             extra_mounts=[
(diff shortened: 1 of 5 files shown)


src/ostorlab/runtimes/lite_local/runtime.py
```diff
@@ -354,9 +370,11 @@ def _inject_assets(self, assets: List[base_asset.Asset]):
```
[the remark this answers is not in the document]
MouadAO: I don't think so because those AgentSettings are the default ones in the yaml file of the inject_asset_agent
```
supported_architectures:
  - linux/arm64/v8
  - linux/x86_64
restart_policy : 'none'
image: images/logo.png
```
you can check it here : https://github.com/Ostorlab/agent_inject_asset/blob/main/ostorlab.yaml

src/ostorlab/runtimes/local/runtime.py
```diff
@@ -546,9 +562,11 @@ def _inject_assets(self, assets: List[base_asset.Asset]):
```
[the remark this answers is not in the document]
MouadAO: I don't think so because those AgentSettings are the default ones in the yaml file of the inject_asset_agent
```
supported_architectures:
  - linux/arm64/v8
  - linux/x86_64
restart_policy : 'none'
image: images/logo.png
```
you can check it here : https://github.com/Ostorlab/agent_inject_asset/blob/main/ostorlab.yaml

src/ostorlab/runtimes/local/runtime.py
```diff
@@ -546,9 +562,11 @@ def _inject_assets(self, assets: List[base_asset.Asset]):
 
         volumes.create_volume(f"asset_{self.name}", contents)
 
-        inject_asset_agent_settings = definitions.AgentSettings(
-            key=ASSET_INJECTION_AGENT_DEFAULT, restart_policy="none"
-        )
+        if inject_asset_agent_settings is None:
```
amat-osto, line 565: Group-supplied inject_asset settings default `restart_policy` to `""`, which `agent_runtime` resolves to `"any"`, so the agent keeps restarting and re-injecting. Can we force `restart_policy="none"` when it is left empty?

src/ostorlab/runtimes/lite_local/runtime.py
```diff
@@ -354,9 +370,11 @@ def _inject_assets(self, assets: List[base_asset.Asset]):
 
         volumes.create_volume(f"asset_{self.name}", contents)
 
-        inject_asset_agent_settings = definitions.AgentSettings(
-            key=ASSET_INJECTION_AGENT_DEFAULT, restart_policy="none"
-        )
+        if inject_asset_agent_settings is None:
```
amat-osto, line 373: Same here: group-supplied inject_asset settings default `restart_policy` to `""` and resolve to `"any"`, so the agent re-injects in a loop. Worth forcing `restart_policy="none"` when it is not set.

src/ostorlab/runtimes/local/runtime.py
MouadAO: I don't think so , my example agent group in the PR didn't restart even though I didn't explicitly added , that is because the inject_asset agent's `ostorlab.yaml` defaults to "none"

```
supported_architectures:
  - linux/arm64/v8
  - linux/x86_64
restart_policy : 'none' # <-- here
image: images/logo.png
```
you can check it here : https://github.com/Ostorlab/agent_inject_asset/blob/main/ostorlab.yaml

src/ostorlab/runtimes/lite_local/runtime.py
MouadAO: some thing:

I don't think so , my example agent group in the PR didn't restart even though I didn't explicitly added , that is because the inject_asset agent's `ostorlab.yaml` defaults to "none"

```
supported_architectures:
  - linux/arm64/v8
  - linux/x86_64
restart_policy : 'none' # <-- here
image: images/logo.png
```
you can check it here : https://github.com/Ostorlab/agent_inject_asset/blob/main/ostorlab.yaml

src/ostorlab/runtimes/local/runtime.py
amat-osto: You're right, the agent's ostorlab.yaml ships restart_policy: 'none' so it runs once as expected. Resolving.

amat-osto (approved):
LGTM

PR #1082, opened 2026-07-07
feature: Update the oxo repo message to support the content url 

zoulatizakaria:
Add RepositoryArchive asset

Adds a new RepositoryArchive asset for scanning a source-code repository
delivered as an archive (tarball/zip), separate from the existing git-based
Repository asset. Repository is left unchanged.

The archive follows the same transport convention as the file assets
(android_apk, ios_ipa, …): the payload can be provided as inline bytes, a
local path, or a remote URL.

Changes

    Proto (v3.asset.file.repository_archive): new repository_archive.proto
    + generated _pb2.py with content (bytes), path, and content_url fields.
    Placed under the file/ asset group.
    Asset (assets/repository_archive.py): new RepositoryArchive with
    optional content / path / content_url.
    CLI (scan run repository-archive): new command with --file / --url
    options, mutually exclusive (mirrors the android-apk command); links the
    created scan to its agent group and assets.
    Registration: wired into cli/scan/run/assets/__init__.py.
    Scanner transport: added repository_archive to the startAgentScan
    oneof (field 30) and regenerated startAgentScan_pb2.py, plus a matching
    branch in scanner/callbacks.py::_extract_assets so the asset actually
    reaches the scanner instead of being dropped.


src/ostorlab/cli/scan/run/assets/repository_archive.py
```diff
@@ -0,0 +1,61 @@
...
+        )
+        raise click.exceptions.Exit(2)
+    if url == () and file == ():
+        console.error(
+            "Command missing either file path or source url of the repository archive."
+        )
+        raise click.exceptions.Exit(2)
+
+    if file != ():
+        for f in file:
```
amat-osto: file -> files (as it is a tuple)

tests/agent/message/proto/v3/asset/repository_archive/repository_archive_pb2_test.py
```diff
@@ -0,0 +1,41 @@
+"""Tests for the repository archive protobuf message definitions and serialization behavior."""
+
+from ostorlab.agent.message.proto.v3.asset.repository_archive import (
+    repository_archive_pb2,
+)
+
+
+def testSerializeAndDeserialize_whenContentUrlSet_returnsEquivalentMessage():
```
amat-osto:
```suggestion
def testSerializeAndDeserialize_whenContentUrlSet_returnsEquivalentMessage() -> None:
```

src/ostorlab/scanner/proto/assets/repository_archive.proto
```diff
@@ -0,0 +1,9 @@
+syntax = "proto2";
+
+package ostorlab.scanner.proto.assets.repository_archive;
+
+message Message {
```
amat-osto, line 5: This scanner-side message is not reachable yet. It is not added to the `oneof asset` in `src/ostorlab/scanner/proto/scan/_location/startAgentScan.proto` (the last asset field there is `repository = 29`), so the scanner transport message has no field to carry a repository archive. Can we add a `repository_archive` field to that oneof and regenerate `startAgentScan_pb2.py` in the same PR? Otherwise this proto is dead code and the asset cannot reach the scanner path.

src/ostorlab/scanner/proto/assets/repository_archive.proto
```diff
@@ -0,0 +1,9 @@
+syntax = "proto2";
+
+package ostorlab.scanner.proto.assets.repository_archive;
```
amat-osto, line 3: Even once this is wired into the `startAgentScan` oneof, `_extract_assets` in `src/ostorlab/scanner/callbacks.py` (around lines 81-120) has no `elif asset_type == "repository_archive"` branch. So `request.WhichOneof("asset")` would resolve to it and then fall through, silently dropping the asset. Can we add the matching branch (mirroring the existing `repository` one) so the scanner actually builds a `RepositoryArchive`?

src/ostorlab/agent/message/proto/v3/asset/file/repository_archive/repository_archive.proto
```diff
@@ -0,0 +1,9 @@
+syntax = "proto2";
+
+package ostorlab.agent.message.proto.v3.asset.repository_archive;
+
+message Message {
+  optional bytes content = 1;
+  optional string path = 2;
+  optional string content_url = 3;
+}
```
mouadostor, line 9: since it is a file i think better to put it in file folder as the other file assets 
`oxo/src/ostorlab/agent/message/proto/v3/asset/file`

zoulatizakaria:
> ## Remaining Outstanding Items
> 
> Based on the latest review round, there are still two critical items that need to be addressed to make the scanner path functional:
> 
>     1. **Missing `oneof asset` field in `startAgentScan.proto`**: The `repository_archive` field has not been added to the `oneof asset` in `src/ostorlab/scanner/proto/scan/_location/startAgentScan.proto`. Without this, the scanner transport message cannot carry a repository archive asset.
> 
>     2. **Missing `_extract_assets` branch in `callbacks.py`**: Even after adding the `oneof` field, `_extract_assets` in `src/ostorlab/scanner/callbacks.py` lacks an `elif asset_type == "repository_archive"` branch to handle the new asset type. This would cause the asset to be silently dropped when `request.WhichOneof("asset")` resolves to it.
> 
> 
> Please address both items in a follow-up commit. The path/package consistency fixes in this round are correct, but the proto will remain dead code on the scanner path until these are resolved.

Could you re-check against the latest commit? The review looks to be on an earlier revision.

src/ostorlab/agent/message/proto/v3/asset/file/repository_archive/repository_archive_pb2.py
```diff
@@ -0,0 +1,36 @@
...
+from google.protobuf import descriptor as _descriptor
+from google.protobuf import descriptor_pool as _descriptor_pool
+from google.protobuf import runtime_version as _runtime_version
+from google.protobuf import symbol_database as _symbol_database
+from google.protobuf.internal import builder as _builder
+_runtime_version.ValidateProtobufRuntimeVersion(
+    _runtime_version.Domain.PUBLIC,
+    6,
+    31,
+    1,
```
mouadostor: couldn't you compile with 6.33.6

mouadostor (approved):

src/ostorlab/agent/message/proto/v3/asset/file/repository_archive/repository_archive_pb2.py
amat-osto: @mouadostor nice comment

amat-osto (approved):

PR #1091, opened 2026-07-13
feature: Support multiple risks per scan

yasser-abassi:
Injects one or more risks through the standard `--assets` asset group.

```bash
oxo scan run -g agents.yaml --assets targets.yaml
```

```yaml
kind: targetGroup
description: Target group with risks
assets:
  risk:
    - severity: HIGH
      description: Server exposed to the internet
      ip:
        host: 8.8.8.8
        mask: 32
    - severity: LOW
      description: Weak TLS configuration
      domain:
        name: example.com
```

Each risk needs `severity` + `description` and may embed at most one target (`ip`, `domain`, `link`, `androidStore`, `iosStore`, `androidApkFile`, `androidAabFile`, `iosFile`). No proto change; the `oxo scan run ... risk` subcommand is unchanged.

## Parsing helpers (`definitions.py`)

- `_parse_risk_asset(entry)` — builds one `Risk` from a YAML risk entry: takes `severity` + `description`, attaches its single embedded target, and rejects invalid input (more than one target, invalid IP, empty file asset).
- `_parse_file_asset(entry)` — resolves a file asset into `(content, path, url)`: reads local bytes when `path` is set, keeps `url` otherwise; returns `None` when neither is usable. Shared by the standalone file loops and risk parsing.
- `_resolve_risk_file_asset(entry, key)` — wraps `_parse_file_asset` for risks: returns the tuple, or raises `ValidationError` when it is `None` so a risk cannot embed an empty file asset.
- `_parse_ip_asset(entry)` — now coerces `mask` to a string. YAML parses `mask: 32` as an int while the proto field is a string, so a masked IP previously failed to serialize with `TypeError: bad argument type for built-in operation`. This also fixes the same latent failure for standalone IP assets.

## Schema (`target_group_schema.json`)

- Added a `riskType` custom type (severity enum + description, plus one optional embedded target: `ip`, `domain`, `link`, `androidStore`, `iosStore`, `androidApkFile`, `androidAabFile`, `iosFile`) and a `risk` array under `assets`.


src/ostorlab/runtimes/definitions.py
```diff
@@ -407,13 +410,109 @@ def from_yaml(cls, group: io.FileIO):
...
+    "link",
+    "androidStore",
+    "iosStore",
+    "androidApkFile",
+    "androidAabFile",
+    "iosFile",
+]
+
+
+def _parse_risk_asset(risk_entry: Dict[str, Any]) -> risk_asset.Risk:
```
yami-ost:
```suggestion
def _parse_risk_asset(risk_entry: dict[str, Any]) -> risk_asset.Risk:
```

src/ostorlab/runtimes/definitions.py
```diff
@@ -407,13 +410,109 @@ def from_yaml(cls, group: io.FileIO):
...
+    would silently drop all but one."""
+    provided_targets = [
+        key for key in _RISK_TARGET_KEYS if risk_entry.get(key) is not None
+    ]
+    if len(provided_targets) > 1:
+        raise validator.ValidationError(
+            f"A risk asset must embed at most one target, got: {', '.join(provided_targets)}."
+        )
+
+    risk_kwargs: Dict[str, Any] = {
```
yami-ost:
```suggestion
    risk_kwargs: dict[str, Any] = {
```

src/ostorlab/runtimes/definitions.py
```diff
@@ -407,13 +412,118 @@ def from_yaml(cls, group: io.FileIO):
...
 
 
+def _parse_file_asset(
+    file_asset: dict[str, Any],
+) -> tuple[bytes | None, str | None, str | None] | None:
+    """Resolve a file asset entry into (content, path, url).
+
+    Returns None when the entry has neither readable content nor a URL, so the
+    caller can skip it (standalone assets) or reject it (embedded risk asset)."""
+    path = file_asset.get("path")
```
Nouri-Anouar, line 438: small bug here: `iosFileType` in the schema defines the key as `paths` (plural) but `_parse_file_asset` reads `file_asset.get("path")` (singular). so if someone passes a local path for an iOS IPA via YAML it will never be picked up, the function returns None, and `_resolve_risk_file_asset` raises a ValidationError. URL-based iOS files work fine, only the local path case is broken. android files are not affected since `androidApkFileType` and `androidAabFileType` both use `path` (singular).

src/ostorlab/runtimes/definitions.py
```diff
@@ -22,12 +23,24 @@
...
 
+_RISK_TARGET_KEYS = [
+    "ip",
+    "domain",
+    "link",
+    "androidStore",
+    "iosStore",
+    "androidApkFile",
+    "androidAabFile",
+    "iosFile",
```
Nouri-Anouar, line 41: the `Risk` dataclass (in `assets/risk.py`) has two fields that are not covered here: `api_schema` and `repository`. they are also missing from the `riskType` schema in `target_group_schema.json`, so there is no way for a user to provide an API schema or a repository target in a risk entry via YAML at all. not sure if that is intentional or just not implemented yet, but worth a comment either way. if it is intentional, it could save confusion to add a note somewhere.

src/ostorlab/runtimes/definitions.py
```diff
@@ -407,13 +412,122 @@ def from_yaml(cls, group: io.FileIO):
...
+    (ip, domain, link, stores and app files). A risk carries a single target
+    because the underlying proto asset is a oneof; embedding more than one
+    would silently drop all but one.
+
+    The ``Risk`` asset also defines ``api_schema`` and ``repository`` targets,
+    but these are intentionally not exposed here since they are not modeled as
+    target group assets anywhere else in ``from_yaml``."""
+    provided_targets = [
+        key for key in _RISK_TARGET_KEYS if risk_entry.get(key) is not None
+    ]
```
m0hamed-ait, line 483: this does not read apiSchema; it's excluded from _RISK_TARGET_KEYS and has no branch, and riskType in `target_group_schema.json` has no apiSchema property. Since the schema sets no `additionalProperties: false`, the validator accepts the YAML silently, so `_parse_risk_asset` produces a targetless Risk and the api-schema target is lost with no error.

src/ostorlab/agent/schema/target_group_schema.json
```diff
@@ -55,15 +55,16 @@
       "description": "[Optional] - Ios app file asset.",
       "type": "object",
       "properties": {
-        "paths": {
+        "path": {
```
3asm: How about cli?

3asm: Why this change?

src/ostorlab/agent/schema/target_group_schema.json
```diff
@@ -234,6 +289,12 @@
           "items" : {
             "$ref": "#/CustomTypes/ticketType"
           }
+        },
+        "risk": {
```
3asm, line 291:
```suggestion
        "risks": {
```

3asm: Isn't this an array?

3asm (dismissed):

src/ostorlab/agent/schema/target_group_schema.json
yasser-abassi: this is a typo

yasser-abassi: this is asset-group YAML only

src/ostorlab/agent/schema/target_group_schema.json
yasser-abassi: following the same naming convention

src/ostorlab/agent/schema/target_group_schema.json
```diff
@@ -157,6 +157,59 @@
...
+          "$ref": "#/CustomTypes/androidAabFileType"
+        },
+        "iosFile": {
+          "$ref": "#/CustomTypes/iosFileType"
+        }
+      },
+      "required": [
+        "severity",
+        "description"
+      ]
```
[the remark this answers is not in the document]
yasser-abassi: Out of scope

src/ostorlab/runtimes/definitions.py
```diff
@@ -422,10 +536,12 @@ def _parse_ip_asset(ip_asset: Dict[str, Any]) -> Optional[base_asset.Asset]:
         logger.info(f"Invalid ip address: {ip_string}")
         return None
 
+    mask = ip_asset.get("mask")
```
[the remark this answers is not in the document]
yasser-abassi: Not valid  ipaddress.ip_address(None) raises ValueError, not TypeError (ip_address() re-raises a plain ValueError for anything it can't parse). So the existing except ValueError already catches it: ip: {mask: 32} returns None from _parse_ip_asset and surfaces as ValidationError: Risk asset has an invalid IP address: None. Adding TypeError would be dead code.

src/ostorlab/runtimes/definitions.py
```diff
@@ -407,13 +412,122 @@ def from_yaml(cls, group: io.FileIO):
...
 
 
+def _parse_file_asset(
+    file_asset: dict[str, Any],
+) -> tuple[bytes | None, str | None, str | None] | None:
+    """Resolve a file asset entry into (content, path, url).
+
+    Returns None when the entry has neither readable content nor a URL, so the
+    caller can skip it (standalone assets) or reject it (embedded risk asset)."""
+    path = file_asset.get("path")
```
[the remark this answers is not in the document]
yasser-abassi: Out of scope

src/ostorlab/runtimes/definitions.py
```diff
@@ -407,13 +412,122 @@ def from_yaml(cls, group: io.FileIO):
...
+        if isinstance(ip_asset, ipv4_asset.IPv4):
+            risk_kwargs["ipv4"] = ip_asset
+        elif isinstance(ip_asset, ipv6_asset.IPv6):
+            risk_kwargs["ipv6"] = ip_asset
+        else:
+            raise validator.ValidationError(
+                f"Risk asset has an invalid IP address: {risk_entry['ip'].get('host')}"
+            )
+
+    if risk_entry.get("domain") is not None:
```
ohachimOs, line 505: domain/link/androidStore/iosStore accept empty sub-objects silently, no ValidationError, unlike ip and the file targets. domain: {} builds DomainName(name=None), crashes on str() the moment anything logs it.

src/ostorlab/runtimes/definitions.py
```diff
@@ -407,13 +412,136 @@ def from_yaml(cls, group: io.FileIO):
...
+        risk_kwargs["domain_name"] = domain_name_asset.DomainName(
+            name=_resolve_risk_target_field(risk_entry, "domain", "name")
+        )
+
+    if risk_entry.get("link") is not None:
+        risk_kwargs["link"] = link_asset.Link(
+            url=_resolve_risk_target_field(risk_entry, "link", "url"),
+            method=_resolve_risk_target_field(risk_entry, "link", "method"),
```
ohachimOs: method still unguarded here, unlike url on the line above. link: {url: x} with no method passes schema and builds Link(method=None).

yasser-abassi: Already fixed

src/ostorlab/runtimes/definitions.py
```diff
@@ -407,13 +412,136 @@ def from_yaml(cls, group: io.FileIO):
...
+    would silently drop all but one.
+
+    The ``Risk`` asset also defines ``api_schema`` and ``repository`` targets,
+    but these are intentionally not exposed here since they are not modeled as
+    target group assets anywhere else in ``from_yaml``."""
+    provided_targets = [
+        key for key in _RISK_TARGET_KEYS if risk_entry.get(key) is not None
+    ]
+    if len(provided_targets) > 1:
+        raise validator.ValidationError(
```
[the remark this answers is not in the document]
yasser-abassi: out-of-scope

src/ostorlab/runtimes/definitions.py
```diff
@@ -407,13 +412,136 @@ def from_yaml(cls, group: io.FileIO):
...
+    if risk_entry.get("iosFile") is not None:
+        content, path, url = _resolve_risk_file_asset(risk_entry, "iosFile")
+        risk_kwargs["ios_ipa"] = ios_ipa_asset.IOSIpa(
+            content=content, path=path, content_url=url
+        )
+
+    return risk_asset.Risk(**risk_kwargs)
+
+
 def _parse_ip_asset(ip_asset: Dict[str, Any]) -> Optional[base_asset.Asset]:
```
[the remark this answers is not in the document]
yasser-abassi: out-of-scope

ohachimOs (dismissed):

src/ostorlab/runtimes/definitions.py
```diff
@@ -407,13 +412,136 @@ def from_yaml(cls, group: io.FileIO):
...
         return cls(
             targets=assets_def,
             name=target_group_def.get("name"),
             description=target_group_def.get("description"),
         )
 
 
+def _parse_file_asset(
+    file_asset: dict[str, Any],
+) -> tuple[bytes | None, str | None, str | None] | None:
```
burogurama: I suggest using named tuples or a dataclass, same for the other parsing functions

src/ostorlab/runtimes/definitions.py
```diff
@@ -407,13 +412,140 @@ def from_yaml(cls, group: io.FileIO):
...
+            assets_def.append(_parse_risk_asset(asset))
+
         return cls(
             targets=assets_def,
             name=target_group_def.get("name"),
             description=target_group_def.get("description"),
         )
 
 
+class ParsedFileAsset(NamedTuple):
```
yami-ost, line 425: If both path and url are provided, content wins from path but url still gets passed through as content_url alongside it. The risk CLI subcommand explicitly rejects providing both (console.error("Provide either --android-apk or --android-apk-url, not both.")), so there's now a strictness mismatch between the two entry points for the same risk feature.

src/ostorlab/runtimes/definitions.py
yasser-abassi: addressed

src/ostorlab/runtimes/definitions.py
yasser-abassi: Out of scope 
this PR doesn't change the file-loading logic. Accepting both path and url (content + content_url together) is the existing standalone-YAML behavior for all file assets on main;

burogurama (approved):

PR #1096, opened 2026-07-15
feature: Add new generic_file asset

zoulatizakaria:
## Add generic file asset

Adds support for scanning arbitrary files via a new generic file asset.

### What's included
- New `generic_file` asset proto (agent message + scanner).
- `GenericFile` asset class with selector `v3.asset.file.generic_file`.
- Wired `generic_file` into the `startAgentScan` asset oneof.
- New CLI command: `oxo scan run generic-file --file <path> | --url <url>`.
- Scanner callback support to decode the asset from a scan request.

A file can be provided inline (`--file`) or by reference (`--url`).


src/ostorlab/agent/message/proto/v3/asset/file/generic_file/generic_file.proto
```diff
@@ -0,0 +1,9 @@
+syntax = "proto2";
```
3asm, line 1: Just file. Why generic why ? ? 😿

3asm (changes requested):

PR #1103, opened 2026-07-20
feature: Add hash_delete to AgentPersistMixin

Nouri-Anouar:
## Summary
- Adds `hash_delete(hash_name, *keys)` to `AgentPersistMixin`, wrapping redis-py's `HDEL`
- Mirrors the existing `hash_add`/`hash_exists`/`hash_get`/`hash_get_all` style

Agents that correlate two paired messages (e.g. an HTTP request and its response, matched by id) park one side in a Redis hash until its partner arrives. Once correlated, there's currently no way to remove that entry -- `delete()` only removes a whole key, not a single hash field -- so entries accumulate forever and a redelivered message re-correlates and re-reports.

## Test plan
- [x] `pytest tests/agent/mixins/agent_persist_mixin_test.py -m docker` (8 passed, including the new `testAgentPersistMixin_whenHashKeyIsDeleted_keyIsRemoved`), run against a real Redis via the existing swarm-service fixture
- [x] `ruff format --check` / `ruff check` clean
- [x] `mypy` -- no new errors (3 pre-existing errors on unrelated lines, confirmed present on `main` too)

### DIFF
--- a/src/ostorlab/agent/mixins/agent_persist_mixin.py
+++ b/src/ostorlab/agent/mixins/agent_persist_mixin.py
@@ -179,6 +179,20 @@ class AgentPersistMixin:
         """
         return self._redis_client.hgetall(hash_name)
 
+    def hash_delete(
+        self, hash_name: Union[bytes, str], *keys: Union[bytes, str]
+    ) -> bool:
+        """Delete one or more keys from the hash hash_name.
+
+        Args:
+            hash_name: Name of the hash.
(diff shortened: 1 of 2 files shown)


Nouri-Anouar:
closing - no longer needed. This was motivated by request/response correlation cleanup in agent_sensitive_info_detector, which got dropped (response scanning removed from that agent's scope). Its own dedup no longer needs per-field deletion.

PR #1106, opened 2026-07-20
feature: adding the multi asset the -a option in the cli

zoulatizakaria:
## Summary

Adds a `multi_asset` section to the target group YAML. Assets declared inside it are bundled and emitted as a **single** `v3.asset.multi_asset` proto message, instead of one message per asset.

## Motivation

`oxo scan run -a targets.yaml` parses the YAML into a flat asset list, and `_inject_assets` writes one `asset.binproto_{i}` / `selector.txt_{i}` pair per asset — so N assets produce N bus messages. The `multi_asset` proto already existed, but nothing could populate it from a YAML.

## Usage

`multi_asset` is a sibling key under `assets`, so bundled and standalone assets coexist in the same file:

```yaml
kind: targetGroup
assets:
  ip:                          # standalone -> one message each
    - host: 1.1.1.1
  domain:
    - name: example.com

  multi_asset:                 # bundled -> ONE message
    ip:
      - host: 10.0.0.1
    ipv4:
      - host: 8.8.8.8
      - host: 192.168.1.0
        mask: 24
    ipv6:
      - host: "2001:4860:4860::8888"
    link:
      - url: https://example.com/login
        method: POST
    file:
      - url: https://host/payload.bin
    repository:
      - repository_url: https://github.com/Ostorlab/oxo
        commit_hash: deadbeef
        provider: GITHUB
    repositoryArchive:
      - url: https://host/source.zip
    androidApkFile:            # oneof -> single object, not a list
      url: https://host/app.apk

### CODE REVIEW

File: src/ostorlab/runtimes/definitions.py [hunk 1]
```diff
@@ -45,6 +54,58 @@
...
+
+_MULTI_ASSET_REPEATED_FILE_CLASSES: dict[str, Any] = {
+    "file": file_asset.File,
+    "repositoryArchive": repository_archive_asset.RepositoryArchive,
+}
+
+_MULTI_ASSET_MOBILE_FILE_CLASSES: dict[str, Any] = {
+    "androidApkFile": android_apk_asset.AndroidApk,
+    "androidAabFile": android_aab_asset.AndroidAab,
+    "iosFile": ios_ipa_asset.IOSIpa,
```
mouadostor:
```suggestion
    "iosIpaFile": ios_ipa_asset.IOSIpa,
```

src/ostorlab/runtimes/definitions.py
```diff
@@ -419,13 +480,145 @@ def from_yaml(cls, group: io.FileIO):
...
+        commit_hash=commit_hash,
+        provider=str(provider),
+    )
+
+
+def _build_multi_asset_file(
+    entry: dict[str, Any], yaml_key: str, asset_class: Any
+) -> base_asset.Asset:
+    """Build a file-backed multi asset target, rejecting entries with no path nor url."""
+    parsed_file = _parse_file_asset(entry)
```
Moonflower1911, line 719: `_parse_file_asset()` only reads `path`, so the case of `multi_asset.iosFile.paths` entry is rejected as missing path/url

src/ostorlab/runtimes/definitions.py
```diff
@@ -419,13 +480,145 @@ def from_yaml(cls, group: io.FileIO):
...
+def _parse_multi_asset(
+    multi_asset_group: dict[str, Any],
+) -> multi_asset_asset.MultiAsset | None:
+    """Build a single multi asset from the target group's multi_asset section.
+
+    Returns None when the section holds no asset, so no empty message is injected."""
+    targets: list[base_asset.Asset] = []
+
+    for yaml_key, asset_class in _MULTI_ASSET_IP_CLASSES.items():
+        for entry in multi_asset_group.get(yaml_key, []):
```
Ayman-ANNA, line 578: i guess here you should use _parse_ip_asset

src/ostorlab/runtimes/definitions.py
```diff
@@ -45,6 +54,58 @@
...
+    "androidAabFile": android_aab_asset.AndroidAab,
+    "iosFile": ios_ipa_asset.IOSIpa,
+    "harmonyosHapFile": harmonyos_hap_asset.HarmonyOSHap,
+    "harmonyosApkFile": harmonyos_apk_asset.HarmonyOSApk,
+    "harmonyosAabFile": harmonyos_aab_asset.HarmonyOSAab,
+    "harmonyosAppFile": harmonyos_app_asset.HarmonyOSApp,
+    "harmonyosRpkFile": harmonyos_rpk_asset.HarmonyOSRpk,
+}
+
+_MULTI_ASSET_STORE_CLASSES: dict[str, tuple[Any, str]] = {
```
Ayman-ANNA: for type safety it is better to use type[base_asset.Asset] instead of Any like in _MULTI_ASSET_MOBILE_FIELDS

src/ostorlab/assets/multi_asset.py
```diff
@@ -81,6 +98,45 @@ def _nested_assets(self) -> list[asset.Asset | None]:
...
+                # __dict__ rather than dataclasses.asdict, which would reintroduce
+                # defaults the asset unset (see Repository.__post_init__).
+                data[key] = [
+                    entry.__dict__ if isinstance(entry, asset.Asset) else entry
+                    for entry in value
+                ]
+            else:
+                data[key] = value
+
+        return bytes(serializer.serialize(self.selector, data).SerializeToString())
```
m0hamed-ait, line 160: SerializeToString() already return a bytes object, you can safely return it without wrapping in bytes()

src/ostorlab/runtimes/definitions.py
```diff
@@ -419,13 +480,145 @@ def from_yaml(cls, group: io.FileIO):
...
+
+    if len(unsupported_targets) > 0:
+        raise validator.ValidationError(
+            f"The multi asset message has no field for the following targets: "
+            f"{', '.join(unsupported_targets)}."
+        )
+
+    bundled_asset = multi_asset_asset.MultiAsset(**multi_asset_kwargs)
+    set_mobile_fields = bundled_asset.set_mobile_asset_fields
+    if len(set_mobile_fields) > 1:
```
m0hamed-ait, line 543: for this same check in multi_asset.py, you return a different Error (ValueError)

src/ostorlab/assets/multi_asset.py
```diff
@@ -81,6 +98,45 @@ def _nested_assets(self) -> list[asset.Asset | None]:
             *self.ipv6s,
         ]
 
+    @property
+    def set_mobile_asset_fields(self) -> list[str]:
```
m0hamed-ait: property named like an action

tests/runtimes/multi_asset_bundling_test.py
```diff
@@ -0,0 +1,202 @@
...
+    ]
+
+
+def testFromYaml_whenIpIpv4AndIpv6_shouldMapToTheirOwnProtoFields() -> None:
+    message = _parse_multi_asset_message("""
+kind: targetGroup
+assets:
+  multi_asset:
+    ip:
+      - host: 10.0.0.1
```
3asm, line 74: ip?

3asm (dismissed):

src/ostorlab/runtimes/definitions.py
```diff
@@ -40,11 +49,59 @@
...
+    file_asset.File: "files",
+    repository_asset.Repository: "repositories",
+    repository_archive_asset.RepositoryArchive: "repository_archives",
+    link_asset.Link: "urls",
+    ip_asset.IP: "ips",
+    ipv4_asset.IPv4: "ipv4s",
+    ipv6_asset.IPv6: "ipv6s",
+}
+
+_MULTI_ASSET_IP_KEYS = ("ip", "ipv4", "ipv6")
```
Nouri-Anouar, line 73: _parse_ip_asset returns IPv4/IPv6 based on the host, never ip.IP. so 'ip' key entries land in ipv4s not ips, and the test expecting message.ips == ["10.0.0.1"] will fail

mouadostor (dismissed):
LGTM

src/ostorlab/agent/schema/target_group_schema.json
```diff
@@ -405,6 +405,104 @@
...
+            "harmonyosStore": {
+              "$ref": "#/CustomTypes/harmonyosStoreType"
+            },
+            "androidApkFile": {
+              "$ref": "#/CustomTypes/androidApkFileType"
+            },
+            "androidAabFile": {
+              "$ref": "#/CustomTypes/androidAabFileType"
+            },
+            "iosIpaFile": {
```
amat-osto: The new `multi_asset` schema accepts `iosIpaFile`, but the existing target-group YAML key for IPA assets is `iosFile`. Because this object has `additionalProperties: false`, users cannot bundle an iOS file with the same key that works for standalone and risk assets. Please support `iosFile` here, or accept both keys and map both to the `ios_ipa` proto field.

src/ostorlab/runtimes/definitions.py
```diff
@@ -58,6 +60,54 @@
...
+
+_MULTI_ASSET_REPEATED_FILE_CLASSES: dict[str, type[base_asset.Asset]] = {
+    "file": file_asset.File,
+    "repositoryArchive": repository_archive_asset.RepositoryArchive,
+}
+
+_MULTI_ASSET_MOBILE_FILE_CLASSES: dict[str, type[base_asset.Asset]] = {
+    "androidApkFile": android_apk_asset.AndroidApk,
```
amat-osto, line 81: This parser mapping only recognizes `iosIpaFile`, while the established target-group key is `iosFile`. Even if the schema is updated, `multi_asset: { iosFile: ... }` will not be parsed unless this mapping also accepts `iosFile`. Please add `"iosFile": ios_ipa_asset.IOSIpa` or otherwise normalize both spellings before building the asset.

amat-osto (commented):
Found two compatibility issues around the iOS file YAML key.

amat-osto (approved):

PR #1107, opened 2026-07-20
feature: Add api_schema support to multi asset

amat-osto:
## What

Lets a multi-asset scan target carry API schemas, so agents declaring `v3.asset.file.api_schema` as an in_selector can be reached through a multi-asset scan.

Every other in_selector of `threat_intelligence` is already carried by `multi_asset` (`ip.v4`, `ip.v6`, `link`, `android.apk`, `android.aab`, `ios.ipa`, `repository`, `file.repository_archive`). `api_schema` was the one gap.

## Changes

| File | Change |
|---|---|
| `scanner/proto/assets/api_schema.proto` | **New.** Mirrors the agent-side `api_schema` message. |
| `scanner/proto/assets/multi_asset.proto` | `repeated api_schemas = 20` |
| `agent/.../multi_asset/multi_asset.proto` | `repeated api_schemas = 19` |
| `assets/multi_asset.py` | `api_schemas` field + included in `_nested_assets()` |
| `cli/scan/run/assets/api_schema.py` | `--schema-type` constrained with `click.Choice` |
| 3 × `_pb2.py` | regenerated with protoc 33.6 |
| 3 test files | api_schema coverage at dataclass, proto and CLI level |

Tags differ between the two trees (20 vs 19) because the scanner proto has a `network` field the agent tree lacks — picked per-tree rather than copied across.

### Why `AndroidMetadata` / `IOSMetadata` / `HarmonyOSMetadata` / `Header` are included

**The metadata oneof carries the identity of the app an API schema was extracted from — which app's traffic produced this endpoint. Agents consume those fields to correlate a schema back to its parent mobile asset, so without them an `api_schema` arriving through a multi-asset scan would lose that association entirely.** This is not speculative surface: `metadata_oneof` is present on **10 of the scanner asset protos and 15 of the agent asset protos**, including `api_schema.proto` and all three of its variants (`graphql`, `openapi`, `wsdl`). Omitting it would have made the scanner `api_schema` the only file-type asset in either tree unable to carry that context, and would have required a second proto change plus coordinated regeneration to add back later.

`Header` / `extra_headers` is included on the same grounds — it mirrors the agent message, and API schema endpoints commonly need auth headers to be reachable.

Field tags are identical to the agent-side message throughout, so the two trees map 1:1 with no conversion table.

### `--schema-type` validation

`--schema-type` was a free-form string documented as "graphql, wsdl, or openapi" but unvalidated, so a typo reached an agent before failing. Now `click.Choice(["graphql", "wsdl", "openapi"], case_sensitive=False)`. Minor behaviour change: previously-accepted unlisted values are now rejected at the CLI.

`schema_type` stays a **string** on the wire rather than becoming an enum — the agent tree uses a string, and proto2 enums are closed, so a schema type added later would be silently dropped into unknown fields by older builds instead of arriving intact.

`26 passed` on the touched test files.


src/ostorlab/cli/scan/run/assets/api_schema.py
```diff
@@ -32,7 +32,8 @@
 )
 @click.option(
     "--schema-type",
-    help="The schema type (graphql, wsdl, or openapi).",
+    type=click.Choice(["graphql", "wsdl", "openapi"], case_sensitive=False),
```
zoulatizakaria: why we need that the validation should happen in parsing the yaml

src/ostorlab/scanner/proto/assets/api_schema.proto
```diff
@@ -0,0 +1,34 @@
+syntax = "proto2";
```
burogurama, line 1: we already have a v3.asset.file.api_schema

please update that one if it does not fit

burogurama (changes requested):

tests/scanner/proto/assets/scanner_multi_asset_pb2_test.py
```diff
@@ -99,6 +99,44 @@ def testMobileAssetOneof_whenHarmonyosFileSet_returnsFileContent() -> None:
         assert getattr(deserialized, field_name).content == b"harmony-bytes"
 
 
+def testSerializeAndDeserialize_whenApiSchemasSet_preservesApiSchemaFields() -> None:
```
1l0c, line 102: missing docstring

tests/scanner/proto/assets/scanner_multi_asset_pb2_test.py
```diff
@@ -99,6 +99,44 @@ def testMobileAssetOneof_whenHarmonyosFileSet_returnsFileContent() -> None:
...
+    assert deserialized.api_schemas[0].endpoint_url == "https://example.com/graphql"
+    assert deserialized.api_schemas[0].schema_type == "graphql"
+    assert deserialized.api_schemas[0].extra_headers[0].name == "Authorization"
+    assert deserialized.api_schemas[0].extra_headers[0].value == "Bearer token"
+    assert (
+        deserialized.api_schemas[0].android_metadata.package_name == "com.example.app"
+    )
+
+
+def testSerializeAndDeserialize_whenMultipleApiSchemas_preservesAllEntries() -> None:
```
1l0c, line 126: +1

tests/assets/multi_asset_test.py
```diff
@@ -57,6 +59,32 @@ def testMultiAsset_whenCreatedWithNestedAssets_storesAllNestedAssets() -> None:
...
+        content_url="https://ostorlab.co/schema.json",
+        schema_type="graphql",
+    )
+
+    asset = multi_asset.MultiAsset(api_schemas=[nested_api_schema])
+
+    assert asset.api_schemas == [nested_api_schema]
+
+
+def testMultiAsset_whenApiSchemaNested_isListedInStringRepresentation() -> None:
```
1l0c, line 74: same here please add to others tests too

src/ostorlab/cli/scan/run/assets/api_schema.py
zoulatizakaria: The cli for that asset is out of scope we want to use the yaml with the -a option no cli

tests/scanner/proto/assets/scanner_multi_asset_pb2_test.py
amat-osto: Why?

tests/scanner/proto/assets/scanner_multi_asset_pb2_test.py
amat-osto: Why?

tests/assets/multi_asset_test.py
amat-osto: Why?

src/ostorlab/scanner/proto/assets/api_schema.proto
amat-osto: That one lives in the agent tree (`ostorlab.agent.message.proto.v3.asset.file.api_schema`), this one is in the scanner tree (`ostorlab.scanner.proto.assets`). Different packages, so `scanner/proto/assets/multi_asset.proto` can't reference the agent message without importing across the two trees.

The scanner tree already duplicates just about every asset for that reason: file, apk, aab, ipa, the harmonyos ones, repository, repository_archive, link, ip, v4, v6, the three stores, domain_name, phone_number, source. api_schema was the only asset without a twin, which is exactly why multi_asset couldn't carry it.

I did look at importing the agent proto straight into the scanner multi_asset (protoc resolves it fine, both are under `src`). I left it alone because no scanner proto imports across today, and the split looks deliberate so the ingress format and the bus format can version independently. Happy to switch if you'd rather kill the duplication, though that makes api_schema the one exception in a 25 file convention.

src/ostorlab/cli/scan/run/assets/api_schema.py
amat-osto: Fair enough, dropped the `click.Choice`.

Worth flagging though: `AssetsDefinition.from_yaml` currently only reads `androidStore`, `androidAabFile`, `androidApkFile`, `iosStore`, `iosFile`, `ip`, `domain`, `link`, `ticket` and `risk`. There's no `apiSchema` key, and no `multiAsset` either, so `-a` can't express an API schema today. We'll wire those into the yaml loader in an upcoming PR.

burogurama (approved):

PR #1111, opened 2026-07-20
fix: Support api schema assets in the target group yaml

amat-osto:
Follow up to the discussion on #1107, where `-a` turned out not to be able to express an API schema at all.

## What

`AssetsDefinition.from_yaml` read ten asset keys (`androidStore`, `androidAabFile`, `androidApkFile`, `iosStore`, `iosFile`, `ip`, `domain`, `link`, `ticket`, `risk`) and had no `apiSchema` among them, so an API schema could only be scanned through `oxo scan run api-schema`, never through `oxo scan run -a assets.yaml`. This adds the missing key.

```yaml
kind: targetGroup
description: Target group with api schemas
assets:
  apiSchema:
    - endpoint_url: https://example.com/graphql
      url: https://example.com/schema.json
      schema_type: graphql
    - endpoint_url: https://example.com/api
      path: ./openapi.json
      schema_type: openapi
    - endpoint_url: https://example.com/soap
```

## Changes

| File | Change |
|---|---|
| `agent/schema/target_group_schema.json` | new `apiSchemaType`, plus `apiSchema` under `assets` |
| `runtimes/definitions.py` | parse `apiSchema` entries into `ApiSchema` assets |
| `tests/runtimes/definitions_test.py` | 4 tests: url form, local path form, missing `endpoint_url`, unsupported `schema_type` |

**`schema_type` is validated by the json schema** (`enum: [graphql, wsdl, openapi]`), so a typo like `grapql` is rejected while the yaml is parsed rather than reaching an agent. This is the placement @zoulatizakaria asked for on #1107; the equivalent `click.Choice` on the CLI command was reverted there.

`endpoint_url` is the only required field, matching the `ApiSchema` dataclass where it is the sole non-defaulted field. An entry with no `path` and no `url` is valid: the schema is then fetched from `endpoint_url` rather than supplied inline.

`61 passed` on the touched test files.

## Notes

- **`multiAsset` is deliberately not included.** It needs the `api_schemas` field added in #1107, so bundling it would block this PR behind that one. It is the natural next step once #1107 lands, and is a larger change since the yaml would need to express the nested assets and the mobile `oneof` semantics.
- Scanning an api schema through `-a` still depends on the agents in the group declaring `v3.asset.file.api_schema` as an in_selector.


burogurama (approved):

zoulatizakaria (approved):
LGTM

PR #1115, opened 2026-07-21
feature: add ASN and network asset messages

Lachhab26:
## Summary
- add `v3.asset.asn` for canonical ASN discovery input
- add `v3.asset.network` for one normalized CIDR per message
- retain `origin_asn` on every network message so ASN-to-network provenance survives downstream processing
- require all fields needed to represent each asset and relationship
- cover IPv4 and IPv6 round trips plus incomplete-message rejection

## Safety boundary
This PR only defines the shared transport contract. A network message represents a discovered range; it does not enumerate or scan individual addresses.

## Verification
- `pytest --no-cov -q tests/agent/message` (94 passed)
- `ruff check .`
- `mypy src/ostorlab/agent/message`
- `python -m build .` (wheel and sdist built; both new protobuf modules included)

## Dependency
First PR in the OS-29817 integration chain. The WHOIS agent, persistence, injector, and reporting-engine changes will consume this contract in separate PRs.


src/ostorlab/agent/message/proto/v3/asset/asn/asn.proto
```diff
@@ -0,0 +1,7 @@
+syntax = "proto2";
+
+package ostorlab.agent.message.proto.v3.asset.asn;
+
+message Message {
+  optional string asn = 1;
```
MouadAO: why it is `optional` ? `asn` it is the only field

Lachhab26: Agreed. I changed `asn` to `required`. I also made `cidr`, `version`, and `origin_asn` required on the Network message because an incomplete message cannot preserve the ASN-to-network relationship, and added negative serialization tests for missing required fields.

PR #1118, opened 2026-07-22
fix: Support MultiAsset Database Persistence in Local Runtime

mouadostor:
## Description
When running a scan with a bundled target group (e.g., oxo scan run -g ag.yaml -a targets.yaml), create_from_assets_definition() in the local database models only checked for standard standalone asset instances (such as ip.IP or link.Link). Because it had no explicit handler for multi_asset.MultiAsset, the bundled container and all of its nested child targets were silently ignored and never persisted to SQLite.
## The Solution
- Updated src/ostorlab/runtimes/local/models/models.py:
- Imported the multi_asset module.
- Added recursive support inside Asset.create_from_assets_definition() to intercept any MultiAsset target, extract its nested child assets (using _nested_assets()), and safely pass them back into the persistence pipeline.
## Verification & Testing
Tested locally against a targets.yaml file containing a multi_asset IP block:
1. Parsed the configuration and started the scan.
2. Verified that all nested targets (e.g. network / IP) are now successfully linked and populated in the local SQLite DB under the correct scan_id:
sqlite> SELECT * FROM asset WHERE scan_id = 712;
201|network|712
## Types of Changes
- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)
- Breaking change (fix or feature that would cause existing functionality to not work as expected)


src/ostorlab/runtimes/local/models/models.py
```diff
@@ -847,7 +848,12 @@ def create_from_assets_definition(
         links: List[Dict[str, str]] = []
         domains: List[Dict[str, str]] = []
         for asset in assets:
-            if (
+            if isinstance(asset, multi_asset.MultiAsset):
+                Asset.create_from_assets_definition(
```
Ayman-ANNA: it is better to avoid recursion

src/ostorlab/runtimes/local/models/models.py
```diff
@@ -889,6 +888,30 @@ def create_from_assets_definition(
         if len(domains) > 0:
             DomainAsset.create(domains=domains, scan_id=scan_id)
 
+    @staticmethod
+    def create_from_assets_definition(
+        assets: Optional[List[base_asset.Asset]], scan_id: Optional[int] = None
```
zoulatizakaria:
```suggestion
        assets: list[base_asset.Asset]|None, scan_id: int|None = None
```

src/ostorlab/runtimes/local/models/models.py
```diff
@@ -889,6 +888,30 @@ def create_from_assets_definition(
...
+        Args:
+            assets: The list of assets to create.
+            scan_id: The scan id.
+        """
+        if assets is None or len(assets) == 0:
+            return
+
+        flattened_assets = []
+        queue = list(assets)
+        while queue:
```
zoulatizakaria: explicit check

zoulatizakaria (dismissed):

src/ostorlab/assets/multi_asset.py
```diff
@@ -97,12 +97,12 @@ class MultiAsset(asset.Asset):
...
         nested_assets = [
             str(nested_asset)
-            for nested_asset in self._nested_assets()
+            for nested_asset in self.nested_assets()
             if nested_asset is not None
         ]
         return f"MultiAsset: [{', '.join(nested_assets)}]"
 
-    def _nested_assets(self) -> list[asset.Asset | None]:
+    def nested_assets(self) -> list[asset.Asset | None]:
```
amat-osto, line 105: why this change?

src/ostorlab/runtimes/definitions.py
```diff
@@ -491,6 +493,28 @@ def from_yaml(cls, group: io.FileIO):
...
+                raise validator.ValidationError(
+                    "Repository requires a non-empty 'repository_url' field."
+                )
+            assets_def.append(_parse_repository_asset(asset))
+
+        for asset in repository_archive_assets:
+            parsed_file = _parse_file_asset(asset)
+            if parsed_file is None:
+                raise validator.ValidationError(
+                    "Repository archive requires either a valid path or a url."
```
amat-osto:
```suggestion
                    "Repository archive requires either a valid path or a content-url."
```

amat-osto (dismissed):

src/ostorlab/assets/multi_asset.py
mouadostor: i did this to fix a bad convention i did before

src/ostorlab/runtimes/local/models/models.py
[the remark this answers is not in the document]
mouadostor: this is not really critical as it is done only once per scan and the existing assets won't be many

1l0c (dismissed):
LGTM

zoulatizakaria (approved):

amat-osto (approved):

PR #1121, opened 2026-07-23
fix: pin the version temporarily

mouadostor:
## The problem

However, we do import opentelemetry-exporter-gcp-trace in src/ostorlab/agent/mixins/agent_open_telemetry_mixin.py:
from opentelemetry.exporter import cloud_trace
The Problem
When Python loads cloud_trace from opentelemetry-exporter-gcp-trace, the library itself immediately executes this import:
from opentelemetry.resourcedetector.gcp_resource_detector import ( ... )
This module belongs to opentelemetry-resourcedetector-gcp. Since opentelemetry-exporter-gcp-trace does not always install it transitively (depending on the environment/Python version), importing cloud_trace will always throw an import error if opentelemetry-resourcedetector-gcp is not present in the environment.
Because this import is at the module level in our OpenTelemetry mixin, it is executed unconditionally every time ostorlab.agent.agent is imported. This completely breaks test collection and prevents any tests from running.


yami-ost (changes requested):
can u pin with a specific version instead of <

PR #1131, opened 2026-07-24
fix: add apischema support to risks

m0hamed-ait:
A risk entry in a target group YAML can embed a target (ip, domain, link, stores, app files, repository…), but apiSchema was never wired in:

- it is missing from _RISK_TARGET_KEYS, so it doesn't count toward the "at most one target" check;
- _parse_risk_asset has no branch reading it;
- riskType in target_group_schema.json doesn't declare it, and the type doesn't set additionalProperties: false.

The three combine into a silent failure: a target group carrying risk: [{severity, description, apiSchema: {...}}] validates cleanly, then produces a Risk with no target at all — the agents receive a risk with an empty proto oneof and no endpoint to test. No exception, no log line.

### DIFF
--- a/src/ostorlab/agent/schema/target_group_schema.json
+++ b/src/ostorlab/agent/schema/target_group_schema.json
@@ -272,6 +272,9 @@
         "link": {
           "$ref": "#/CustomTypes/linkType"
         },
+        "apiSchema": {
+          "$ref": "#/CustomTypes/apiSchemaType"
+        },
         "androidStore": {
           "$ref": "#/CustomTypes/androidStoreType"
         },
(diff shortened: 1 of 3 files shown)


MouadAO (approved):

MouadAO (approved):

MouadAO (approved):

MouadAO (approved):

zoulatizakaria (approved):

src/ostorlab/agent/schema/target_group_schema.json
yami-ost, line 1: riskType still has no additionalProperties: false even though your own pr description calls this out as one of the 3 causes

m0hamed-ait: just explained that one of them either additionalProperties: true or adding api_schema is a solution.

PR #1139, opened 2026-07-28
fix: make psutil a core dependency instead of a scanner extra

burogurama:
Moves `psutil` from the optional `scanner` extra into the core `install_requires`.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -62,6 +62,7 @@ install_requires =
     py-ubjson
     nats-py
     pyaxmlparser
+    psutil
 
 [options.packages.find]
 where = src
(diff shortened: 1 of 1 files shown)


src/ostorlab/scanner/resource_checker.py
```diff
@@ -28,6 +26,16 @@ def can_run_scan(
         logger.warning("No resource requirements configured for scan %s", scan_key)
         return False
 
+    try:
+        import psutil
```
3asm: Simply add psutil the setup.cfg deps.

3asm (dismissed):

m0hamed-ait (approved):

yasser-abassi (approved):

PR #1156, opened 2026-08-04
fix: add API key support to ScannerStateReporter and update tests

Ayman-ANNA:
## Issue

OXO successfully fetched the scanner configuration and started scans, but reporting the scanner state through `addScannerState` failed with `401 Unauthorized`.

## Cause

The Reporting Engine API key was used to fetch `ScannerConfig`, but it was not passed to `ScannerStateReporter`. Consequently, the reporter created an unauthenticated API client in the scanner child process.

This is separate from `ScannerConfig.api_key`, which authenticates scan polling against the Scanning Engine.

## Fix

Propagate the existing Reporting Engine organization API key to `ScannerStateReporter` and use it when calling `addScannerState`.

No additional API key or infrastructure configuration is required.

### DIFF
--- a/src/ostorlab/cli/scanner/scanner.py
+++ b/src/ostorlab/cli/scanner/scanner.py
@@ -125,7 +125,10 @@ def scanner(
     import daemon as dm
 
     state_reporter = scanner_state_reporter.ScannerStateReporter(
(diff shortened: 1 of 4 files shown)


1l0c (approved):

ohachimOs (approved):

yami-ost (approved):
LGTM but address the bot's comments

PR #1157, opened 2026-08-04
fix: add support for GCP logging credentials in scanner processes

Ayman-ANNA:
## Issue

Scans launched through `oxo scanner` were running successfully, but agent logs were not appearing in Google Cloud Logging.

## Cause

The scanner command loaded the GCP logging credential, but did not propagate it through the scanner worker and scan handler to `LocalRuntime`.

As a result, agent containers were created without the `GCP_LOGGING_CREDENTIAL` environment variable and could not initialize Google Cloud Logging.

## Fix

Propagate the existing GCP logging credential through:

`scanner CLI → scanner worker → ScanHandler → callbacks.start_scan → LocalRuntime → agent containers`

No changes were required to the existing local-runtime logging implementation.

### DIFF
--- a/src/ostorlab/cli/scanner/scanner.py
+++ b/src/ostorlab/cli/scanner/scanner.py
@@ -117,6 +117,7 @@ def scanner(
         raise click.exceptions.Exit(2)
 
     api_key = config_manager.ConfigurationManager().api_key or ctx.obj.get("api_key")
+    gcp_logging_credential = ctx.obj.get("gcp_logging_credential")
     scanner_log_file = log_file if persist_logs is True else None
     scanner_log_level = getattr(logging, log_level.upper())
     _configure_file_logging(scanner_log_file, scanner_log_level)
@@ -139,6 +140,7 @@ def scanner(
                 state_reporter,
                 scanner_log_file,
                 scanner_log_level,
+                gcp_logging_credential,
             ),
         )
         process.start()
@@ -159,6 +161,7 @@ def start_scanner(
     state_reporter: scanner_state_reporter.ScannerStateReporter,
     log_file: str | None = None,
     log_level: int = logging.INFO,
+    gcp_logging_credential: str | None = None,
 ) -> None:
     """Run subscription to nats in event loop.
 
@@ -168,6 +171,7 @@ def start_scanner(
         state_reporter: instance responsible for reporting the scanner state.
         log_file: Optional path to persist scanner logs.
         log_level: Logging level used for persisted scanner logs.
+        gcp_logging_credential: GCP Logging JSON credentials for agent containers.
     """
     _configure_file_logging(log_file, log_level)
     if api_key is None:
@@ -184,4 +188,5 @@ def start_scanner(
         api_key=api_key,
         scanner_id=scanner_id,
         state_reporter=state_reporter,
+        gcp_logging_credential=gcp_logging_credential,
     )
--- a/src/ostorlab/cli/scanner/scanner.py
+++ b/src/ostorlab/cli/scanner/scanner.py
@@ -117,6 +117,7 @@ def scanner(
         raise click.exceptions.Exit(2)
 
     api_key = config_manager.ConfigurationManager().api_key or ctx.obj.get("api_key")
+    gcp_logging_credential = ctx.obj.get("gcp_logging_credential")
     scanner_log_file = log_file if persist_logs is True else None
     scanner_log_level = getattr(logging, log_level.upper())
     _configure_file_logging(scanner_log_file, scanner_log_level)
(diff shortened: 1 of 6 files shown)


yami-ost (dismissed):

tests/scanner/scan_handler_test.py
```diff
@@ -198,6 +198,35 @@ def testTriggerScanWithRollback_whenStartScanFails_rollsBack(
...
+        scanner_id="GGBD-DJJD-DKJK-DJDD",
+        hostname="test-host",
+        ip="192.168.0.1",
+    )
+    scan_handler_instance = scan_handler.ScanHandler(
+        state_reporter=state_reporter,
+        gcp_logging_credential="gcp-credential",
+    )
+
+    result = scan_handler_instance._trigger_scan_with_rollback(
```
yasser-abassi: tests a private method

yasser-abassi (dismissed):
small comments

tests/scanner/callbacks_test.py
```diff
@@ -70,6 +70,45 @@ def testExtractAssets_whenApkAsset_shouldReturnCorrectAsset(
...
+            "key": "agentgroup/ostorlab/agent_group42",
+            "agents": [{"key": "agent/ostorlab/dummy"}],
+        },
+        "asset": {
+            "__typename": "AndroidApkAssetType",
+            "content": base64.b64encode(b"dummy_apk").decode(),
+        },
+    }
+    mocker.patch("ostorlab.scanner.callbacks._connect_containers_registry")
+    mocker.patch("ostorlab.scanner.callbacks._update_state_reporter")
```
mouadostor: we shouldn't mock private functions as it is forbidden by our code convention.
try to avoid that as you did in the whole file

src/ostorlab/cli/scanner/scanner.py
```diff
@@ -117,6 +117,7 @@ def scanner(
         raise click.exceptions.Exit(2)
 
     api_key = config_manager.ConfigurationManager().api_key or ctx.obj.get("api_key")
+    gcp_logging_credential = ctx.obj.get("gcp_logging_credential")
```
amat-osto, line 120: rootcli.py only sets ctx.obj["gcp_logging_credential"] after the google.cloud.logging import succeeds. Without that extra installed, it stays None and containers get no env var, so the fix silently no-ops. Could we assign it right after open(...)?

amat-osto (dismissed):

src/ostorlab/scanner/callbacks.py
```diff
@@ -370,7 +372,10 @@ def start_scan(
         )
 
         runtime_instance = registry.select_runtime(
-            runtime_type="local", scan_id=str(scan_id), run_default_agents=False
+            runtime_type="local",
+            scan_id=str(scan_id),
```
Oussama-Bakri, line 376: good PR, however I have a different question about hard coding the runtime_type in case of providing a scanner, can you create a different PR and instead use the provided runtime, if no runtime is provided, then use "local" runtime by default.

tests/cli/scanner/scanner_test.py
```diff
@@ -103,6 +103,38 @@ def testScannerCommandInvocation_whenApiKeyProvided_passesItToStateReporter(
     )
 
 
+@pytest.mark.skipif(sys.platform == "win32", reason="does not run on windows")
+def testScannerCommandInvocation_whenGcpCredentialProvided_forwardsItToProcess(
+    mocker: plugin.MockerFixture,
+) -> None:
+    """Forward GCP logging credentials to every scanner process."""
+    mocker.patch.object(scanner_cli, "_configure_file_logging")
```
Oussama-Bakri, line 111: we don't mock, call... private functions in tests, please fix

src/ostorlab/scanner/callbacks.py
Ayman-ANNA: agree i will do it in another pr

yami-ost (dismissed):

Oussama-Bakri (approved):

Oussama-Bakri (dismissed):
Remove all private functions in tests

tests/scanner/callbacks_test.py
```diff
@@ -70,6 +70,45 @@ def testExtractAssets_whenApkAsset_shouldReturnCorrectAsset(
...
+        "agentGroup": {
+            "key": "agentgroup/ostorlab/agent_group42",
+            "agents": [{"key": "agent/ostorlab/dummy"}],
+        },
+        "asset": {
+            "__typename": "AndroidApkAssetType",
+            "content": base64.b64encode(b"dummy_apk").decode(),
+        },
+    }
+    mocker.patch("ostorlab.scanner.callbacks._connect_containers_registry")
```
amat-osto: private function, don't think we should mock it.

amat-osto (dismissed):

yami-ost (approved):

PR #1160, opened 2026-08-06
feature: add v3.asset.asn message contract

Lachhab26:
## Summary

- Add the shared v3.asset.asn protobuf message.
- Carry asn plus optional description, matching the payload emitted by agent_asn_discovery.
- Cover prefixed and numeric-only ASN payloads through Message.from_data and Message.from_raw.

This intentionally does not change v3.asset.network. Existing network messages already represent one CIDR with a singleton ips entry containing host, mask, and version.

## Verification

- ASN serializer tests: 2 passed
- Full v3 asset-proto tests: 40 passed
- Ruff format and check: passed
- Mypy in the repository typing environment: passed
- Wheel and sdist build: passed
- Generated protobuf version: 6.33.6

## Integration order

This contract must be merged and released before agent_asn_discovery streams v3.asset.asn and before agent_whois_ip consumes it.

Supersedes #1148.

Ticket: https://report.ostorlab.co/o/os/remediation/tickets/os-29817


src/ostorlab/agent/message/proto/v3/asset/asn/asn.proto
```diff
@@ -0,0 +1,8 @@
+syntax = "proto2";
+
+package ostorlab.agent.message.proto.v3.asset.asn;
+
+message Message {
+  optional string asn = 1;
+  optional string description = 2;
```
3asm: How will we get description if we scan an ASN. This should not be part of an ASN in asset.

3asm (changes requested):

3asm (approved):

PR #1164, opened 2026-08-11
fix: Pass api_key to fetch agent details since the ConfigManager api key value is lost when starting a new process.

amine3:
This PR adds support for authenticating API requests using an api_key in the get_details function of agent_fetcher.py. 

This is needed since we pass the APi_key as a parameter and should not rely on the singleton config manager. The config manager was initiated in the main process, and the forked process lost the values.
 
  Changes
   1. API Support (src/ostorlab/cli/agent_fetcher.py)
      - Added an optional api_key: str | None = None parameter to the get_details function.
      - Updated the authentication logic within get_details to check if api_key is provided when the
        manager is not authenticated. If an api_key is passed, it instantiates AuthenticatedAPIRunner
        with that key; otherwise, it falls back to the PublicAPIRunner.

   2. Unit Tests (tests/cli/test_agent_fetcher.py)
      - Added the unit test testGetDetails_whenApiKeyProvided_buildsAuthenticatedRunnerWithApiKey to
        verify that passing an api_key to get_details correctly instantiates AuthenticatedAPIRunner with
        the specified key and executes the details query successfully.

### DIFF
--- a/src/ostorlab/cli/install_agent.py
+++ b/src/ostorlab/cli/install_agent.py
@@ -161,7 +161,7 @@ def install(
         click Exit exception with status code 2 when the docker image does not exist.
     """
 
-    agent_details = agent_fetcher.get_details(agent_key)
+    agent_details = agent_fetcher.get_details(agent_key, api_key=api_key)
     agent_docker_location = agent_details["dockerLocation"]
     if agent_docker_location is None or not agent_details.get("versions", {}).get(
         "versions", []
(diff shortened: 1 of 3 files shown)


src/ostorlab/cli/install_agent.py
```diff
@@ -161,7 +161,7 @@ def install(
         click Exit exception with status code 2 when the docker image does not exist.
     """
 
-    agent_details = agent_fetcher.get_details(agent_key)
+    agent_details = agent_fetcher.get_details(agent_key, api_key=api_key)
```
3asm: Named arg for all.

3asm (dismissed):

Oussama-Bakri (approved):

PR #1175, opened 2026-08-26
fix: keep caller logging during migrations

yasser-abassi:
The on-prem scanner stops logging entirely at its first migration.

`Database` runs alembic in-process, and `env.py` calls `fileConfig(alembic.ini)`, which is process-global: it replaces the root handlers, drops the root level to `WARN` and disables the already existing loggers. From that point the process is silent for the rest of its life, so the scan lifecycle messages shipped in #1169 have no logger left to emit through.

Production shows this precisely. The last entry ever recorded for the on-prem scanner is the migration itself, after which the worker keeps scanning but never logs again:

```
2026-08-19T11:00:07.783745Z  pid 30  Context impl SQLiteImpl.
2026-08-19T11:00:07.785318Z  pid 30  Will assume non-transactional DDL.
```

The scanner container has no volume on `~/.ostorlab`, so every restart starts from an empty database and the migration always runs, on the first scan.

`env.py` now honors a `configure_logger` config attribute and `Database` opts out. The alembic CLI still configures its own logging.

```mermaid
flowchart LR
    A["Database migrate"] --> B["env.py fileConfig"]
    B -- "before" --> C["root handlers replaced, loggers disabled"]
    C --> D["process silent for its lifetime"]
    B -- "after" --> E{"configure_logger?"}
    E -- "CLI: true" --> F["fileConfig"]
    E -- "embedded: false" --> G["caller logging untouched"]
```


mouadostor (approved):
LGTM

amine3 (approved):

PR #1179, opened 2026-08-27
fix: support apiSchema scan when URL has apiSchema field

amine3:
When a UrlAssetType carries an apiSchema field, extract it as an ApiSchema asset with the first URL as endpoint_url and the apiSchema value as content_url. Also add api_schema handling in risk target keyword args.

This ensures the scanner can run an API schema scan when the URL asset has an associated schema file, instead of falling back to a plain Link scan.

Changes:
- `_extract_assets`: handle `apiSchema` field on `UrlAssetType` to produce `ApiSchema` assets
- `_build_risk_kwargs`: add `ApiSchema` target asset mapping
- Tests: cover UrlAssetType with/without apiSchema extraction


src/ostorlab/scanner/callbacks.py
```diff
@@ -279,6 +281,15 @@ def _extract_assets(asset_data: dict[str, Any]) -> list[asset.Asset]:
             for network in kwargs.get("networks") or []
         ]
     elif typename == "UrlAssetType":
+        api_schema_url = kwargs.get("apiSchema")
+        if api_schema_url is not None and str(api_schema_url).strip() != "":
+            endpoint_url = kwargs.get("urls", [""])[0] if kwargs.get("urls") else ""
```
amat-osto: Just make this check explicit.

amat-osto (dismissed):
LGTM

src/ostorlab/scanner/callbacks.py
```diff
@@ -279,6 +281,17 @@ def _extract_assets(asset_data: dict[str, Any]) -> list[asset.Asset]:
             for network in kwargs.get("networks") or []
         ]
     elif typename == "UrlAssetType":
+        api_schema_url = kwargs.get("apiSchema")
+        if api_schema_url is not None and str(api_schema_url).strip() != "":
+            endpoint_url = (
+                kwargs.get("urls", [""])[0] if kwargs.get("urls") is not None else ""
```
yami-ost, line 287: if apiSchema is set and urls has more than one entry, only urls[0] becomes endpoint_url and the rest are silently dropped, no log or warning. before this change all of them would become Link assets and get scanned

yami-ost, line 287: the latest push changed this to `is not None`, which changes behavior for urls=[]: kwargs.get("urls", [""])[0] now runs since "urls" key exists, but get returns the real value [] not the [""] default, so this is [][0], IndexError. urls missing/None still falls through fine to endpoint_url="". _extract_api_schema_assets skips and logs "Multi asset api schema holds no endpoint url" for this exact case instead of crashing or leaving an empty endpoint

src/ostorlab/scanner/callbacks.py
```diff
@@ -279,6 +281,17 @@ def _extract_assets(asset_data: dict[str, Any]) -> list[asset.Asset]:
             for network in kwargs.get("networks") or []
         ]
     elif typename == "UrlAssetType":
```
yami-ost, line 283: this duplicates _extract_api_schema_assets's construction logic instead of calling it, which is why its missing that functions endpoint validation

src/ostorlab/scanner/callbacks.py
yami-ost, line 287: the [""] default on the first .get() is dead code, this branch only runs when kwargs.get("urls") is not None so the key is already known present, get never falls back to the default. thats also why urls=[] crashes instead of hitting a fallback, see the other comment

src/ostorlab/scanner/callbacks.py
amine3: No, in the case of an API schema, we persist it as a URL with schema content.

PR #1187, opened 2026-09-02
fix: derive scanner concurrency limit from running universes

burogurama:
## What?

The scanner daemon decides whether it may take another scan by remembering the
one scan it started, in a local variable:

```python
scan_id = None
while True:
    if scan_id is not None and self._is_scan_running(scan_id=scan_id) is True:
        sleep; continue
    ...
```

`_is_scan_running` only ever asks Docker about that one remembered id. This
replaces it with `_count_running_universes`, which counts the distinct
`ostorlab.universe` label values across the host's services, and gates the loop
on that count against a new `max_concurrent_scans` limit. The limit is the
existing `--parallel` value, plumbed through `start_scanner` and
`start_scan_loop` to `ScanHandler`.

On a Docker error the count now returns the limit, so the host is treated as
full rather than free.

## Why?

The remembered `scan_id` is process-local, so it is lost whenever the daemon
container is replaced. Universes are Swarm services and keep running across
that restart, so the new worker starts with `scan_id = None`, skips the guard
entirely, and reserves another scan within a few seconds of booting — on top of
whatever the host is already running.

On one production host this ran as a loop: an agent process grew to ~25 GB on a
31 GB machine, the kernel OOM killer fired and took the daemon container with
it, Swarm restarted the daemon, and the fresh worker claimed another scan,
which added more memory pressure. That host recorded 29 OOM kills and 12 daemon
restarts in 24 hours. Another reached 19 restarts, 24 reservations and 5
concurrent universes, with a load average of 255 on 8 cores, and stopped
accepting SSH.

Counting the host's own running universes makes the limit survive a restart,
because it is derived from the machine's state rather than from process memory.

## Changed files

- `src/ostorlab/scanner/scan_handler.py`
- `src/ostorlab/cli/scanner/scanner.py`
- `tests/scanner/scan_handler_test.py`
- `tests/cli/scanner/scanner_test.py`

### DIFF
--- a/src/ostorlab/cli/scanner/scanner.py
+++ b/src/ostorlab/cli/scanner/scanner.py
@@ -211,6 +211,7 @@ def scanner(
                 scanner_log_file,
                 scanner_log_level,
                 gcp_logging_credential,
+                nb_parallel_scans,
             ),
         )
         process.start()
@@ -232,6 +233,7 @@ def start_scanner(
     log_file: str | None = None,
     log_level: int = logging.INFO,
     gcp_logging_credential: str | None = None,
+    max_concurrent_scans: int = 1,
 ) -> None:
     """Run subscription to nats in event loop.
 
@@ -242,6 +244,7 @@ def start_scanner(
         log_file: Optional path to persist scanner logs.
         log_level: Logging level used for persisted scanner logs.
         gcp_logging_credential: GCP Logging JSON credentials for agent containers.
+        max_concurrent_scans: Number of universes the host may run at once.
     """
     _configure_file_logging(log_file, log_level)
     _configure_gcp_logging(gcp_logging_credential, scanner_id)
@@ -260,4 +263,5 @@ def start_scanner(
         scanner_id=scanner_id,
         state_reporter=state_reporter,
         gcp_logging_credential=gcp_logging_credential,
+        max_concurrent_scans=max_concurrent_scans,
     )
--- a/src/ostorlab/scanner/scan_handler.py
+++ b/src/ostorlab/scanner/scan_handler.py
@@ -54,7 +54,16 @@ class ScanHandler:
 
         while True:
             running_universes = self._count_running_universes()
-            if running_universes >= self._max_concurrent_scans:
+            if running_universes > self._max_concurrent_scans:
+                logger.error(
+                    "Host is running %s universe(s) over a limit of %s. Sleeping...",
+                    running_universes,
+                    self._max_concurrent_scans,
+                )
+                time.sleep(WAIT_CHECK_MESSAGES.seconds)
+                continue
+
+            if running_universes == self._max_concurrent_scans:
                 logger.debug(
                     "Host is running %s universe(s) for a limit of %s. Sleeping...",
                     running_universes,
--- a/src/ostorlab/scanner/scan_handler.py
+++ b/src/ostorlab/scanner/scan_handler.py
@@ -21,6 +21,7 @@ from ostorlab.utils import scanner_state_reporter
 logger = logging.getLogger(__name__)
 
 WAIT_CHECK_MESSAGES = datetime.timedelta(seconds=5)
+UNIVERSE_LABEL = "ostorlab.universe"
 
 
 class ScanHandler:
@@ -32,10 +33,12 @@ class ScanHandler:
         scan_resource_requirements: dict[str, scanner_conf.ScanResourceRequirements]
         | None = None,
         gcp_logging_credential: str | None = None,
+        max_concurrent_scans: int = 1,
     ):
         self._state_reporter = state_reporter
         self._scan_resource_requirements = scan_resource_requirements or {}
         self._gcp_logging_credential = gcp_logging_credential
+        self._max_concurrent_scans = max_concurrent_scans
         self._docker_client = docker.from_env()
 
     def close(self) -> None:
@@ -47,19 +50,18 @@ class ScanHandler:
         api_key: str | None = None,
     ) -> None:
         """Scan handler method responsible for fetching messages from the API and triggering the scan."""
-        scan_id = None
-
         logger.info("Starting main API polling loop.")
 
         while True:
-            if scan_id is not None and self._is_scan_running(scan_id=scan_id) is True:
-                logger.debug("Scan %s is still running. Sleeping...", scan_id)
+            running_universes = self._count_running_universes()
+            if running_universes >= self._max_concurrent_scans:
+                logger.debug(
+                    "Host is running %s universe(s) for a limit of %s. Sleeping...",
+                    running_universes,
+                    self._max_concurrent_scans,
+                )
                 time.sleep(WAIT_CHECK_MESSAGES.seconds)
                 continue
-
-            if scan_id is not None:
-                logger.debug("Scan %s has finished. Ready for next.", scan_id)
-            scan_id = None
 
             scans_list = self._fetch_available_scans(runner=runner)
             if scans_list is None or len(scans_list) == 0:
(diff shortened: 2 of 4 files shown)


MouadAO (approved):

src/ostorlab/scanner/scan_handler.py
```diff
@@ -32,10 +33,12 @@ def __init__(
         scan_resource_requirements: dict[str, scanner_conf.ScanResourceRequirements]
         | None = None,
         gcp_logging_credential: str | None = None,
+        max_concurrent_scans: int = 1,
     ):
         self._state_reporter = state_reporter
         self._scan_resource_requirements = scan_resource_requirements or {}
         self._gcp_logging_credential = gcp_logging_credential
+        self._max_concurrent_scans = max_concurrent_scans
```
amine3, line 41: This implementation is a duplicate of  the logic defined in scanner.py for _ in range(nb_parallel_scans):

Here we need to pass the value nb_parallel_scans to check the count.

amine3 (changes requested):

src/ostorlab/scanner/scan_handler.py
```diff
@@ -47,19 +50,27 @@ def handle_messages(
...
         """Scan handler method responsible for fetching messages from the API and triggering the scan."""
-        scan_id = None
-
         logger.info("Starting main API polling loop.")
 
         while True:
-            if scan_id is not None and self._is_scan_running(scan_id=scan_id) is True:
-                logger.debug("Scan %s is still running. Sleeping...", scan_id)
+            running_universes = self._count_running_universes()
+            if running_universes > self._max_concurrent_scans:
```
yasser-abassi, line 57: same branch twice, collapse to >=

src/ostorlab/scanner/scan_handler.py
```diff
@@ -47,19 +50,27 @@ def handle_messages(
...
+                logger.error(
+                    "Host is running %s universe(s) over a limit of %s. Sleeping...",
```
yasser-abassi, line 59: over-capacity branch untested. add a case above the limit

src/ostorlab/scanner/scan_handler.py
```diff
@@ -239,28 +250,41 @@ def _rollback_scan_state(
...
+        Returns:
+            The number of distinct `ostorlab.universe` label values across the
+            host services. On a Docker error, the concurrency limit is returned
+            so the caller treats the host as full.
+        """
         try:
-            scan_services: list[services.Service] = self._docker_client.services.list(
-                filters={"label": f"ostorlab.universe={scan_id!s}"}
+            universe_services: list[services.Service] = (
+                self._docker_client.services.list(
```
yasser-abassi: on a swarm manager this lists cluster-wide services, not this host?

tests/scanner/scan_handler_test.py
```diff
@@ -29,8 +30,8 @@ def testHandleMessages_whenApiKeyProvided_forwardsApiKeyToStartScan(
     )
     mocker.patch.object(
         scan_handler.ScanHandler,
-        "_is_scan_running",
-        return_value=True,
+        "_count_running_universes",
```
yasser-abassi, line 33: mocks a private method. drive it through handle_messages instead

tests/scanner/scan_handler_test.py
```diff
@@ -52,18 +53,49 @@ def testHandleMessages_whenApiKeyProvided_forwardsApiKeyToStartScan(
...
+    mocker: plugin.MockerFixture,
+) -> None:
+    """_count_running_universes should count distinct universes, not services."""
     state_reporter = scanner_state_reporter.ScannerStateReporter(
         scanner_id="GGBD-DJJD-DKJK-DJDD",
         hostname="test-host",
         ip="192.168.0.1",
     )
     scan_handler_instance = scan_handler.ScanHandler(state_reporter=state_reporter)
+    scan_handler_instance._docker_client = mocker.MagicMock()
```
yasser-abassi, line 66: constructor already ran real docker.from_env. patch it before building

src/ostorlab/cli/scanner/scanner.py
```diff
@@ -211,6 +211,7 @@ def scanner(
                 scanner_log_file,
                 scanner_log_level,
                 gcp_logging_credential,
+                nb_parallel_scans,
```
yami-ost, line 214: because --parallel N spawns N start_scanner processes above and passes each max_concurrent_scans=N here. so all N daemons gate on the same _count_running_universes() with no coordination: they wake from the same 5s sleep, read the same count, and if its < N they all reserve at once, overshooting by up to N-1 every settle cycle. the old per-process scan_id capped each at exactly 1. pass 1 here instead (N processes, each self limited, still restart safe), or run one daemon with a semaphore of N.

src/ostorlab/scanner/scan_handler.py
yami-ost, line 263: this counts every ostorlab.universe service on the host, but _max_concurrent_scans comes from this scanner's --parallel. two scanner daemons on one box each with --parallel 2 will run 2 scans total, not 2 each, since each sees the other's universes. the per-scanner limit is now coupled to unrelated universe workloads on the machine.

src/ostorlab/scanner/scan_handler.py
```diff
@@ -47,19 +50,27 @@ def handle_messages(
...
-        scan_id = None
-
         logger.info("Starting main API polling loop.")
 
         while True:
-            if scan_id is not None and self._is_scan_running(scan_id=scan_id) is True:
-                logger.debug("Scan %s is still running. Sleeping...", scan_id)
+            running_universes = self._count_running_universes()
+            if running_universes > self._max_concurrent_scans:
+                logger.error(
```
yami-ost, line 58: this logs at ERROR every 5s for as long as the host is over capacity (a second daemon, leftover universes, another scanner's parallelism). pre PR this wait path was DEBUG. it floods error logs and trips ERROR level alerting while the scanner is just waiting for a slot. keep it DEBUG, or log once on transition into the over-capacity state.

src/ostorlab/scanner/scan_handler.py
```diff
@@ -47,19 +50,27 @@ def handle_messages(
...
+            if running_universes == self._max_concurrent_scans:
+                logger.debug(
+                    "Host is running %s universe(s) for a limit of %s. Sleeping...",
+                    running_universes,
+                    self._max_concurrent_scans,
+                )
+                time.sleep(WAIT_CHECK_MESSAGES.seconds)
+                continue
 
             scans_list = self._fetch_available_scans(runner=runner)
```
yami-ost, line 75: further down this loop: after _trigger_scan_with_rollback returns a non-None scan_id there is no time.sleep, only the scan_id is None path sleeps. so on success the loop comes straight back here to re-count and re-fetch. the scan just started takes minutes before its universe shows in services.list, so this near tight loop can reserve several back to back before the count catches up. the old code got away with the missing sleep because the scan_id guard caught the next iteration.

src/ostorlab/scanner/scan_handler.py
```diff
@@ -47,19 +50,27 @@ def handle_messages(
...
+                    running_universes,
+                    self._max_concurrent_scans,
+                )
                 time.sleep(WAIT_CHECK_MESSAGES.seconds)
                 continue
 
-            if scan_id is not None:
-                logger.debug("Scan %s has finished. Ready for next.", scan_id)
-            scan_id = None
+            if running_universes == self._max_concurrent_scans:
```
yami-ost, line 66: this and the > max block above are the same sleep/continue, differing only in log level and text. could be one >= max check with the level chosen by > vs ==. two copies will drift.

src/ostorlab/scanner/scan_handler.py
```diff
@@ -47,19 +50,27 @@ def handle_messages(
...
     ) -> None:
         """Scan handler method responsible for fetching messages from the API and triggering the scan."""
-        scan_id = None
-
         logger.info("Starting main API polling loop.")
 
         while True:
-            if scan_id is not None and self._is_scan_running(scan_id=scan_id) is True:
-                logger.debug("Scan %s is still running. Sleeping...", scan_id)
+            running_universes = self._count_running_universes()
```
yasser-abassi, line 56: _count_running_universes counts every service carrying the ostorlab.universe label with no liveness filter, so services left behind by a crashed, failed, or half-removed scan (0/1 replicas, tasks all dead) count forever. Once stale universes reach the limit the host stops picking up any scan permanently, and the > branch spams logger.error

amine3 (approved):

src/ostorlab/cli/scanner/scanner.py
burogurama: That's not the bug we re fixing to fix in this PR. read the description

src/ostorlab/scanner/scan_handler.py
burogurama: The host should never be over capacity, if it is then it is an error, and there is no clear way around the repeated logging, we need to see it

src/ostorlab/scanner/scan_handler.py
burogurama: I like them separate, it is not a competition about writing the least number of lines

src/ostorlab/scanner/scan_handler.py
burogurama: Wrong, the function returns after launching the services

src/ostorlab/scanner/scan_handler.py
burogurama: As I said before, this is not the bug this PR fixes

src/ostorlab/scanner/scan_handler.py
burogurama: overcapacity is not the subject of this fix

tests/scanner/scan_handler_test.py
burogurama: I don t understand

src/ostorlab/scanner/scan_handler.py
burogurama: Not the PR job to fix this

PR #1191, opened 2026-09-10
fix: escape rich markup in console output

m0hamed-ait:
### Issue

Printing any text holding brackets, like a regex or a URL, raises a `MarkupError` because rich parses it as markup:

```
console.info(f"Injecting asset: {asset}")
rich.errors.MarkupError: closing tag '[/Pattern]' at position 127 doesn't match any open tag
```

On the scanner, the error escapes `_inject_assets`, the scan is rolled back to `not_started` and picked up again seconds later, looping.

### Fix

- Escape the dynamic text in `Console.info/error/warning/success/status`, with an `is_markup` flag for the few messages that hold intentional markup.
- Escape the string cells of `Console.table`, and return the styled risk ratings, progress and assets as `Text`, so the styles still render while the untrusted values stay escaped.

### DIFF
--- a/tests/cli/test_console.py
+++ b/tests/cli/test_console.py
@@ -1,6 +1,8 @@
 from pytest_mock import plugin
+from rich import table as rich_table
 
 from ostorlab.cli import console as cli_console
+from ostorlab.utils import styles
 
 
 def testConsoleInfo_whenLoggerSet_shouldLogMessages(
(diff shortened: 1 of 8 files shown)


yami-ost (approved):

Lachhab26 (approved):

tests/cli/test_console.py
```diff
@@ -73,3 +75,83 @@ def testConsoleWarning_whenLoggerNotSet_shouldNotLogMessages(
...
+def testConsoleErrorWarningSuccess_whenTextHoldsUnbalancedBrackets_shouldNotRaiseMarkupError() -> (
+    None
+):
+    text = "closing tag '[/Pattern]' at position 127"
+    console = cli_console.Console()
+
+    console.error(text)
+    console.warning(text)
+    console.success(text)
+    console.status(text)
```
sohaibharraoui, line 96: testConsoleInfo_whenTextHoldsUnbalancedBrackets_shouldNotRaiseMarkupError and testConsoleErrorWarningSuccess assert only that no exception is raised, not that the escaped text is actually passed to the console. Patch console._console.print with mocker and assert print_mock.call_args receives the escaped brackets (\\[/Pattern\\]) like testConsoleInfo_whenIsMarkupSet does.

PR #1196, opened 2026-09-17
fix: resolve unpinned agent version before installing on the scanner daemon

yami-ost:
The on-prem scanner daemon's `_install_agents` passes an agent's version straight through to `install_agent.install()` without resolving it first. When a scan profile leaves an agent's version unset (the normal case -- most agent group definitions never pin one), `install()` correctly resolves the version via the experimental-aware `agent_fetcher.get_details(use_experimental=...)` and pulls the right image. But the runtime later starts the container via `AgentSettings.container_image`, which calls `agent_fetcher.get_container_image(key, self.version)` -- and since `self.version` is still `None` at that point, that function ignores the server entirely and picks the semver-`max()` of whatever images happen to be cached locally, with no awareness of the experimental flag. A plain release always outranks a prerelease of the same core version under semver precedence, so an org with `use_experimental_agents=True` still ends up running the stable build even after the correct experimental version was freshly pulled.

This mirrors the existing, already-correct pattern in `cli/scan/run/run.py`'s `_install_agents_with_retry`: when `agent.version is None`, resolve it via `agent_fetcher.get_details` and write it back onto the agent object *before* calling `install()`, so the same `AgentSettings` instance the runtime later reads for `container_image` carries the resolved version instead of `None`.

Confirmed against a real repro: an org with the experimental channel enabled had `v2.5.0-alpha` freshly pulled onto the scanner (per `docker images`), but the actually running container used stable `v2.5.0` -- exactly the symptom this fixes.

**Tests:** two new cases in `tests/scanner/callbacks_test.py` -- resolving via `get_details` when no version is pinned, and falling through gracefully to `install()` with `version=None` when the agent isn't found on the store.


tests/scanner/callbacks_test.py
```diff
@@ -923,6 +924,94 @@ def testStartScan_whenUseExperimentalAgentsIsAbsent_forwardsFalseToInstallAgent(
...
+            "agents": [
+                {"key": "agent/ostorlab/unknown_agent"},
+            ],
+        },
+        "asset": {
+            "__typename": "AndroidApkAssetType",
+            "content": base64.b64encode(b"dummy_apk").decode(),
+        },
+    }
+    mocker.patch("ostorlab.scanner.callbacks._connect_containers_registry")
```
mouadostor: try not to test private functions or mock them

mouadostor (dismissed):
LGTM

amine3 (dismissed):

tests/scanner/callbacks_test.py
yami-ost: Fixed.

src/ostorlab/scanner/callbacks.py
```diff
@@ -86,6 +86,19 @@ def _install_agents(
...
         runtime_instance.install(docker_client=docker_client)
         for agent in agents:
+            if agent.version is None:
+                try:
+                    agent_details = agent_fetcher.get_details(
+                        agent.key,
+                        use_experimental=use_experimental_agents,
+                        api_key=api_key,
+                    )
+                    agent.version = agent_details["versions"]["versions"][0]["version"]
```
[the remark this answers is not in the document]
yami-ost: Fixed.

ohachimOs (approved):

